PR #20, opened 2023-10-26
chore: break circular dependency for crates.io

cbgbt:
Crates with circular dependencies like the one expressed between bottlerocket-settings-sdk and bottlerocket-template-helper must be released to crates.io via a bootstrap process which prevents cargo from enforcing that dependencies are already present on crates.io.

A similar process was used to bootstrap crates like serde, which has the same dependency layout as these crates.

I will revert this commit once we have published `0.1.0` to crates.io.


cbgbt:
Instead of publishing this to crates.io, I'm planning to take this as a git dependency in the Bottlerocket repo so we can collect feedback from real world usage.

PR #37, opened 2024-06-10
feat: add defaults helper

bcressey:
**Issue #, if available:**
N/A

**Description of changes:**
This is extracted from [the storewolf build script](https://github.com/bottlerocket-os/bottlerocket/blob/3538b0c0270852e10a082602ada6c92b7fc9d516/sources/api/storewolf/build.rs), so it can be used as the build script for other crates that want to merge TOML files.

**Terms of contribution:**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,4 +4,5 @@ resolver = "2"
 members = [
     "bottlerocket-settings-sdk",
     "bottlerocket-template-helper",
+    "bottlerocket-defaults-helper",
 ]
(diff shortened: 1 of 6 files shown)


cbgbt (approved):
Do you envision that we will remove this from storewolf's build script after `core-kit` is separated from the main bottlerocket repo?

bcressey:
> Do you envision that we will remove this from storewolf's build script after core-kit is separated from the main bottlerocket repo?

It'll go away from storewolf before then, as part of the refactor in https://github.com/bottlerocket-os/bottlerocket/pull/4038.

webern (approved):

PR #46, opened 2024-07-31
feat: add settings models for bootstrap commands

piyush-jena:
*Issue #, if available:*

*Description of changes:*
This PR contains the following changes:
1. settings sdk changes for supporting settings.bootstrap-commands in bottlerocket.
2. Modified deny.toml to get a successful build. BSD-2 license and multiple versions of crates were allowed. Currently the repository doesn't have a Cargo.lock file which means there is no control over the dependency tree. 

The change in deny.toml will be done in a separate commit after the code review and before merging.

### DIFF
--- a/bottlerocket-settings-models/modeled-types/src/shared.rs
+++ b/bottlerocket-settings-models/modeled-types/src/shared.rs
@@ -385,7 +385,7 @@ mod test_etc_hosts_entries {
 /// character in user-facing identifiers. It stores the original form and makes it accessible
 /// through standard traits. Its purpose is to validate input for identifiers like container names
 /// that might be used to create files/directories.
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct Identifier {
     inner: String,
 }
@@ -945,50 +945,125 @@ string_impls_for!(Lockdown, "Lockdown");
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+/// ApiclientCommand represents a valid Bootstrap Command. It stores the command as a vector of
+/// strings and ensures that the first argument is apiclient.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Default, Serialize)]
+pub struct ApiclientCommand(Vec<String>);
+
+impl ApiclientCommand {
+    pub fn get_command_and_args(&self) -> (&str, &[String]) {
+        self.0
+            .split_first()
+            .map(|(command, rest)| (command.as_str(), rest))
+            .unwrap_or_default()
+    }
+}
+
+// Custom deserializer added to enforce rules to make sure the command is valid.
+impl<'de> serde::Deserialize<'de> for ApiclientCommand {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let original: Vec<String> = serde::Deserialize::deserialize(deserializer)?;
+        Self::try_from(original).map_err(|e| {
+            <D::Error as serde::de::Error>::custom(format!(
+                "Unable to deserialize into ApiclientCommand: {}",
+                e
+            ))
+        })
+    }
+}
+
+impl TryFrom<Vec<String>> for ApiclientCommand {
+    type Error = error::Error;
+
+    fn try_from(input: Vec<String>) -> std::result::Result<Self, error::Error> {
+        let first_word = input.first().map(String::as_str);
+        ensure!(
+            matches!(first_word, Some("apiclient")),
+            error::InvalidCommandSnafu { input },
+        );
+
+        Ok(ApiclientCommand(input))
+    }
+}
+
+#[cfg(test)]
+mod test_valid_apiclient_command {
+    use super::ApiclientCommand;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn valid_apiclient_command() {
+        assert!(ApiclientCommand::try_from(vec![
+            "apiclient".to_string(),
+            "set".to_string(),
+            "motd=helloworld".to_string()
+        ])
+        .is_ok());
+    }
+
+    #[test]
+    fn empty_apiclient_command() {
+        assert!(ApiclientCommand::try_from(Vec::new()).is_err());
+    }
+
+    #[test]
+    fn invalid_apiclient_command() {
+        assert!(ApiclientCommand::try_from(vec![
+            "/usr/bin/touch".to_string(),
+            "helloworld".to_string()
+        ])
+        .is_err());
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
-pub struct BootstrapContainerMode {
+pub struct BootstrapMode {
     inner: String,
 }
 
-impl TryFrom<&str> for BootstrapContainerMode {
+impl TryFrom<&str> for BootstrapMode {
     type Error = error::Error;
 
     fn try_from(input: &str) -> Result<Self, error::Error> {
         ensure!(
             matches!(input, "off" | "once" | "always"),
-            error::InvalidBootstrapContainerModeSnafu { input }
+            error::InvalidBootstrapModeSnafu { input }
         );
-        Ok(BootstrapContainerMode {
+        Ok(BootstrapMode {
             inner: input.to_string(),
         })
     }
 }
 
-impl Default for BootstrapContainerMode {
+impl Default for BootstrapMode {
     fn default() -> Self {
-        BootstrapContainerMode {
+        BootstrapMode {
             inner: "off".to_string(),
         }
     }
 }
 
-string_impls_for!(BootstrapContainerMode, "BootstrapContainerMode");
+string_impls_for!(BootstrapMode, "BootstrapMode");
 
 #[cfg(test)]
 mod test_valid_container_mode {
-    use super::BootstrapContainerMode;
+    use super::BootstrapMode;
     use std::convert::TryFrom;
 
     #[test]
     fn valid_container_mode() {
         for ok in &["off", "once", "always"] {
-            assert!(BootstrapContainerMode::try_from(*ok).is_ok());
+            assert!(BootstrapMode::try_from(*ok).is_ok());
         }
     }
 
     #[test]
     fn invalid_container_mode() {
-        assert!(BootstrapContainerMode::try_from("invalid").is_err());
+        assert!(BootstrapMode::try_from("invalid").is_err());
     }
 }
 
--- a/bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
+++ b/bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
@@ -1,0 +1,148 @@
+//! Settings related to bootstrap commands.
+use bottlerocket_model_derive::model;
+use bottlerocket_modeled_types::{ApiclientCommand, BootstrapMode, Identifier};
+use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::{collections::BTreeMap, convert::Infallible};
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct BootstrapCommandsSettingsV1 {
+    pub bootstrap_commands: BTreeMap<Identifier, BootstrapCommand>,
+}
+
+// Custom serializer/deserializer added to maintain backwards
+// compatibility with models created prior to settings extensions.
+impl Serialize for BootstrapCommandsSettingsV1 {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.bootstrap_commands.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for BootstrapCommandsSettingsV1 {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let bootstrap_commands = BTreeMap::deserialize(deserializer)?;
+        Ok(Self { bootstrap_commands })
+    }
+}
+
+#[model(impl_default = true)]
+struct BootstrapCommand {
+    commands: Vec<ApiclientCommand>,
+    mode: BootstrapMode,
+    essential: bool,
+}
+
+impl SettingsModel for BootstrapCommandsSettingsV1 {
+    type PartialKind = Self;
+    type ErrorKind = Infallible;
+
+    fn get_version() -> &'static str {
+        "v1"
+    }
+
+    fn set(_current_value: Option<Self>, _target: Self) -> Result<()> {
+        // Set anything that parses as BootstrapCommandsSettingsV1.
+        Ok(())
+    }
+
+    fn generate(
+        existing_partial: Option<Self::PartialKind>,
+        _dependent_settings: Option<serde_json::Value>,
+    ) -> Result<GenerateResult<Self::PartialKind, Self>> {
+        Ok(GenerateResult::Complete(
+            existing_partial.unwrap_or_default(),
+        ))
+    }
+
+    fn validate(_value: Self, _validated_settings: Option<serde_json::Value>) -> Result<()> {
+        // Validate anything that parses as BootstrapCommandsSettingsV1.
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test_bootstrap_command {
+    use super::*;
+    use serde_json::json;
+
+    #[test]
+    fn test_generate_bootstrap_command_settings() {
+        let generated = BootstrapCommandsSettingsV1::generate(None, None).unwrap();
+
+        assert_eq!(
+            generated,
+            GenerateResult::Complete(BootstrapCommandsSettingsV1 {
+                bootstrap_commands: BTreeMap::new(),
+            })
+        )
+    }
+
+    #[test]
+    fn test_serde_bootstrap_command() {
+        let test_json = json!({
+            "mybootstrap": {
+                "commands": [ ["apiclient", "motd=hello"], ],
+                "mode": "once",
+                "essential": true,
+            }
+        });
+
+        let bootstrap_commands: BootstrapCommandsSettingsV1 =
+            serde_json::from_value(test_json.clone()).unwrap();
+
+        let mut expected_bootstrap_commands: BTreeMap<Identifier, BootstrapCommand> =
+            BTreeMap::new();
+        expected_bootstrap_commands.insert(
+            Identifier::try_from("mybootstrap").unwrap(),
+            BootstrapCommand {
+                commands: Some(vec![ApiclientCommand::try_from(vec![
+                    "apiclient".to_string(),
+                    "motd=hello".to_string(),
+                ])
+                .unwrap()]),
+                mode: Some(BootstrapMode::try_from("once").unwrap()),
+                essential: Some(true),
+            },
+        );
+
+        assert_eq!(
+            bootstrap_commands,
+            BootstrapCommandsSettingsV1 {
+                bootstrap_commands: expected_bootstrap_commands
+            }
+        );
+
+        let serialized_json: serde_json::Value = serde_json::to_string(&bootstrap_commands)
+            .map(|s| serde_json::from_str(&s).unwrap())
+            .unwrap();
+
+        assert_eq!(serialized_json, test_json);
+    }
+
+    #[test]
+    fn test_serde_invalid_bootstrap_command() {
+        let test_err_json = json!({
+            "mybootstrap1": {
+                "commands": [ ["/usr/bin/touch", "helloworld"], ],
+                "mode": "once",
+                "essential": true,
+            }
+        });
+
+        let bootstrap_commands_err: std::result::Result<
+            BootstrapCommandsSettingsV1,
+            serde_json::Error,
+        > = serde_json::from_value(test_err_json.clone());
+
+        // This has invalid command. It should fail.
+        assert!(bootstrap_commands_err.is_err());
+    }
+}
+
+type Result<T> = std::result::Result<T, Infallible>;
--- a/bottlerocket-settings-models/settings-models/src/lib.rs
+++ b/bottlerocket-settings-models/settings-models/src/lib.rs
@@ -27,6 +27,7 @@ pub use bottlerocket_string_impls_for as string_impls_for;
 pub use crate::boot::BootSettingsV1;
 pub use settings_extension_autoscaling::{self, AutoScalingSettingsV1};
 pub use settings_extension_aws::{self, AwsSettingsV1};
+pub use settings_extension_bootstrap_commands::{self, BootstrapCommandsSettingsV1};
 pub use settings_extension_bootstrap_containers::{self, BootstrapContainersSettingsV1};
 pub use settings_extension_cloudformation::{self, CloudFormationSettingsV1};
 pub use settings_extension_container_registry::{self, RegistrySettingsV1};
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
     # These will eventually live in the kit workspaces that own the related software
     "bottlerocket-settings-models/settings-extensions/autoscaling",
     "bottlerocket-settings-models/settings-extensions/aws",
+    "bottlerocket-settings-models/settings-extensions/bootstrap-commands",
     "bottlerocket-settings-models/settings-extensions/bootstrap-containers",
     "bottlerocket-settings-models/settings-extensions/cloudformation",
     "bottlerocket-settings-models/settings-extensions/container-registry",
@@ -63,6 +64,7 @@ bottlerocket-string-impls-for = { path = "./bottlerocket-settings-models/string-
 ## Settings Extensions
 settings-extension-autoscaling = { path = "./bottlerocket-settings-models/settings-extensions/autoscaling", version = "0.1" }
 settings-extension-aws = { path = "./bottlerocket-settings-models/settings-extensions/aws", version = "0.1" }
+settings-extension-bootstrap-commands = { path = "./bottlerocket-settings-models/settings-extensions/bootstrap-commands", version = "0.1" }
 settings-extension-bootstrap-containers = { path = "./bottlerocket-settings-models/settings-extensions/bootstrap-containers", version = "0.1" }
 settings-extension-cloudformation = { path = "./bottlerocket-settings-models/settings-extensions/cloudformation", version = "0.1" }
 settings-extension-container-registry = { path = "./bottlerocket-settings-models/settings-extensions/container-registry", version = "0.1" }
--- a/bottlerocket-settings-models/modeled-types/src/lib.rs
+++ b/bottlerocket-settings-models/modeled-types/src/lib.rs
@@ -71,8 +71,8 @@ pub mod error {
         #[snafu(display("Invalid Kubernetes authentication mode '{}'", input))]
         InvalidAuthenticationMode { input: String },
 
-        #[snafu(display("Invalid bootstrap container mode '{}'", input))]
-        InvalidBootstrapContainerMode { input: String },
+        #[snafu(display("Invalid bootstrap mode '{}'", input))]
+        InvalidBootstrapMode { input: String },
 
         #[snafu(display("Given invalid cluster name '{}': {}", name, msg))]
         InvalidClusterName { name: String, msg: String },
@@ -85,6 +85,9 @@ pub mod error {
 
         #[snafu(display("Invalid Linux lockdown mode '{}'", input))]
         InvalidLockdown { input: String },
+
+        #[snafu(display("Invalid Bottlerocket API Command '{:?}'", input))]
+        InvalidCommand { input: Vec<String> },
 
         #[snafu(display("Invalid sysctl key '{}': {}", input, msg))]
         InvalidSysctlKey { input: String, msg: String },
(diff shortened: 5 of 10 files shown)


bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,108 @@ string_impls_for!(Lockdown, "Lockdown");
...
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// BootstrapCommandMode determines when the specified set of commands should be executed during
+/// the boot process of a system: 'always' runs them on every boot, 'once' runs them only on the
+/// first boot, and 'off' disables their execution.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct BootstrapCommandMode {
+    inner: String,
+}
```
cbgbt: This seems like it should probably be an enum, since there's a small set of discrete valid inputs.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
```diff
@@ -0,0 +1,128 @@
+//! Settings related to bootstrap commands.
+use bottlerocket_model_derive::model;
+use bottlerocket_modeled_types::{ApiCommand, BootstrapCommandMode, Identifier};
+use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::{collections::HashMap, convert::Infallible};
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct BootstrapCommandsSettingsV1 {
```
cbgbt, line 9: Do you think you could document here on what bootstrap commands are, and perhaps the intended order of bootstrap command execution? (or e.g. if that order is undefined)

cbgbt (commented):
As a tip, you can run `make build` locally to emulate what the CI will do in this repo. Seems like there are a few changes it wants you to make (`cargo fmt`).

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: Updated in the latest iteration.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
piyush-jena: Added comments about BootstrapCommandsSettingsV1 and the order of execution of bootstrap commands. In a separate comment below I have added the details of the order of execution of the commands within a bootstrap command.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,92 @@ string_impls_for!(Lockdown, "Lockdown");
...
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// BootstrapCommandMode determines when the specified set of commands should be executed during
+/// the boot process of a system: 'always' runs them on every boot, 'once' runs them only on the
+/// first boot, and 'off' disables their execution.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Default, Scalar)]
+#[serde(rename_all = "lowercase")]
+pub enum BootstrapCommandMode {
+    #[default]
+    #[serde(rename = "off")]
```
cbgbt: nit: these individual `rename`s shouldn't be necessary since you've used `rename_all`.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,92 @@ string_impls_for!(Lockdown, "Lockdown");
...
+pub struct ApiCommand(pub Vec<String>);
+
+impl TryFrom<Vec<String>> for ApiCommand {
+    type Error = error::Error;
+
+    fn try_from(input: Vec<String>) -> Result<Self, error::Error> {
+        ensure!(
+            input.is_empty() || (input[0] == "apiclient"),
+            error::InvalidApiCommandSnafu { input }
+        );
```
cbgbt: This code is totally fine as-is, I'm not requesting changes. Just commentary.

As a stylistic point of preference, I like to avoid code that could potentially panic, even if the panic conditions are checked ahead of time. I like this because it makes it harder for modifications by another programmer in the future to accidentally invalidate the preconditions and result in a panic. Getting the type system to do this for you is a great benefit if you can.

e.g. in this case, `input[0]` will panic for an empty input; however, you've protected from this earlier in the conditional.

Using `Vec::get` or `Vec::first` returns an `Option<&T>` instead of panicking, so one way you could potentially write this to avoid checked panic preconditions could be something like:

```rust
let binary = input.first().map(String::as_str);
ensure!(
    matches!(binary, None | Some("apiclient")),
    error::InvalidApiCommandSnafu { input }
);
```

cbgbt (approved):
Nice work!

bottlerocket-settings-models/modeled-types/src/shared.rs
arnaldo2792: I wonder if there is a way to enforce this style with clippy. It will be nice to have that rule that we standardize the style.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: Updated

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: I like this suggestion and updated it.

cbgbt: @arnaldo2792 I opened #48 which should enable this on a workspace level.

cbgbt (approved):

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,90 @@ string_impls_for!(Lockdown, "Lockdown");
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Default, Serialize, Deserialize)]
+pub struct ApiCommand(pub Vec<String>);
+
+impl TryFrom<Vec<String>> for ApiCommand {
+    type Error = error::Error;
+
+    fn try_from(input: Vec<String>) -> Result<Self, error::Error> {
+        let first_word = input.first().map(String::as_str);
+        ensure!(
+            matches!(first_word, None | Some("apiclient")),
```
arnaldo2792: Does this mean match either `None` or `apiclient`? Shouldn't this always be `apiclient`?

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,90 @@ string_impls_for!(Lockdown, "Lockdown");
...
+    #[test]
+    fn valid_api_command() {
+        assert!(
+            ApiCommand::try_from(vec!["apiclient".to_string(), "motd=helloworld".to_string()])
+                .is_ok()
+        );
+    }
+
+    #[test]
+    fn empty_api_command() {
```
arnaldo2792: Yeah, I don't think we should allow empty vectors, if no command was given we should error out.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,90 @@ string_impls_for!(Lockdown, "Lockdown");
...
+
+#[cfg(test)]
+mod test_bootstrap_command_mode {
+    use super::BootstrapCommandMode;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn valid_bootstrap_command_mode() {
+        for mode in &["off", "once", "always"] {
+            BootstrapCommandMode::try_from(*mode).unwrap();
```
arnaldo2792:
```suggestion
            assert!(BootstrapCommandMode::try_from(*mode).is_ok());
```

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,90 @@ string_impls_for!(Lockdown, "Lockdown");
...
+        }
+    }
+
+    #[test]
+    fn invalid_bootstrap_command_mode() {
+        for mode in &["", "sometimes", "bad", &"a".repeat(64)] {
+            BootstrapCommandMode::try_from(*mode).unwrap_err();
```
arnaldo2792:
```suggestion
            assert!(BootstrapCommandMode::try_from(*mode).is_err());
```

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+                "commands": [ ["apiclient", "motd=hello"], ],
+                "mode": "once",
+                "essential": true,
+            }
+        });
+
+        let test_json_str = test_json.to_string();
+
+        let bootstrap_commands: BootstrapCommandsSettingsV1 =
+            serde_json::from_str(&test_json_str).unwrap();
```
arnaldo2792: You could do:

```suggestion
        let bootstrap_commands: BootstrapCommandsSettingsV1 =
            serde_json::from_value(&test_json).unwrap();
```

And remolve `test_json_str`

deny.toml
```diff
@@ -7,7 +7,7 @@ confidence-threshold = 0.93
 # Commented license types are allowed but not currently used
 allow = [
     "Apache-2.0",
-    # "BSD-2-Clause",
```
ytsssun: Is this needed?

piyush-jena: Yes. Due to the absence of Cargo.lock, we don't have control over the dependency tree which causes `make build` to fails. This commit is also a part of another PR - #48

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: fixed

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: fixed

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: fixed

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
piyush-jena: switched it to `serde_json::from_value(test_json.clone()).unwrap();` because `serde_json::from_value` accepts `Value`.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: fixed

deny.toml
```diff
@@ -7,7 +7,7 @@ confidence-threshold = 0.93
...
+    "BSD-2-Clause",
```
bcressey: What dependency are you pulling in that's new and licensed `BSD-2-Clause`? This is a surprising change.

deny.toml
```diff
@@ -44,7 +44,6 @@ exceptions = [
 
 [bans]
 # Deny multiple versions or wildcard dependencies.
-multiple-versions = "deny"
```
bcressey: Don't change this.

If you did, you should also remove the comment that no longer applies. But don't.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,90 @@ string_impls_for!(Lockdown, "Lockdown");
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+/// ApiCommand represents a valid Bottlerocket API Command. It stores the command as a vector of
+/// strings and ensures that the first argument is apiclient.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Default, Serialize, Deserialize)]
+pub struct ApiCommand(pub Vec<String>);
+
+impl TryFrom<Vec<String>> for ApiCommand {
+    type Error = error::Error;
```
bcressey: I don't think we should default to the shared types crate for types that aren't likely to see much reuse (like this one). They can live in the extension instead.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,90 @@ string_impls_for!(Lockdown, "Lockdown");
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+/// ApiCommand represents a valid Bottlerocket API Command. It stores the command as a vector of
+/// strings and ensures that the first argument is apiclient.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Default, Serialize, Deserialize)]
+pub struct ApiCommand(pub Vec<String>);
```
bcressey:
```suggestion
pub struct ApiclientCommand(pub Vec<String>);
```

bcressey: Why does the inner type need to be `pub`?
```suggestion
pub struct ApiCommand(Vec<String>);
```

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,6 +945,90 @@ string_impls_for!(Lockdown, "Lockdown");
...
+/// the boot process of a system: 'always' runs them on every boot, 'once' runs them only on the
+/// first boot, and 'off' disables their execution.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Default, Scalar)]
+#[serde(rename_all = "lowercase")]
+pub enum BootstrapCommandMode {
+    #[default]
+    Off,
+    Once,
+    Always,
+}
```
bcressey: I don't think we have other enums in these types. I'm not confident that this will serialize and deserialize properly.

You could think about renaming `BootstrapContainerMode` to `BootstrapMode` and reusing that type for both bootstrap containers and commands.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// BootstrapCommandsSettingsV1 represents how a BootstrapCommand lives as a Bottlerocket Setting.
+/// It can be set by either user-data or through Bottlerocket API. Lets take an example
+/// [settings.bootstrap-commands."001-test-bootstrap-commands"]
+/// commands = [["apiclient", "set", "motd=helloworld1"]]
+/// mode = "always"
+/// essential = true
+/// Here, the name (Identifier) is "001-test-bootstrap-commands" and commands, mode and essential
+/// form the BootstrapCommand object.
+/// The Bootstrap Commands are executed in lexicographical order of their names (Identifier).
```
bcressey: This is an atypical comment for this code base. You are specifying the end usage (`settings.bootstrap-commands`) but that isn't really applied or enforced here. I don't think it clarifies this code to introduce the wider context.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    where
+        D: Deserializer<'de>,
+    {
+        let bootstrap_commands = HashMap::deserialize(deserializer)?;
+        Ok(Self { bootstrap_commands })
+    }
+}
+
+/// BootstrapCommand represents the set of settings within a Bootstrap Command
+/// The commands within a BootstrapCommand are run serially one after the other.
```
bcressey: This property - serial execution - is not something that this code enforces or cares about, which makes the comment distracting and unhelpful.

deny.toml
piyush-jena: None of my changes require this "directly". The issue is that due to the absence of Cargo.lock, one of the dependencies is updating itself to the latest version and demanding this and the multiple-versions change for a successful build. Sean was forced to cherry pick the commit in his PR #48.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: I agree but it felt more general than BootstrapContainerMode placed here. I should move this one to settings-extensions and change BootstrapContainerMode to BootstrapMode and reuse it. It would then justify BootstrapContainerMode's presence in shared crate.

deny.toml
cbgbt: We don't have a Cargo.lock file for the settings SDK since it contains only library crates. This makes it impossible to predict `cargo-deny` errors on licensing and multiple versions, but the multiple versions are definitely the "flappiest," and don't generally have a route to resolving them.

We could introduce a lockfile to control when the CI will break. The benefit would be that we could see if introducing a new dependency introduces undue crate duplication downstream without randomly breaking CI. I advocated for disabling the `multiple-versions` check under the assumption that maintaining the lockfile wasn't worth the benefits.

bottlerocket-settings-models/modeled-types/src/shared.rs
cbgbt: We do use enums this way [elsewhere](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/291f07d214372809b29570db0589784432e2c4bc/bottlerocket-settings-models/modeled-types/src/ecs.rs#L177). The `scalar-derive` macro handles implementing the rest of the string-like traits.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
cbgbt: I advocated to add these because I was missing this piece of the mental puzzle, but you're right: it's not warranted here since the model does not dictate it be used this way.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: Updated. ApiclientCommand is now moved to settings-extensions/bootstrap-commands/src/lib.rs. I would like another review of that file.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: Had kept it public to be able to access the values. In the current revision, I have added get_args function to fetch the arguments.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: I have tested it with Unit tests and end to end testing of bootstrap commands feature and it works.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
piyush-jena: Had added it on request but I agree it deviates from the repository standard and adds additional context not relevant here. Removed it now.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: Updated.

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: We can continue having this discussion but for now I have repurposed BootstrapContainerMode by renaming it to BootstrapMode.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
yeazelm: FWIW I don't feel like the comments you put in there helped me with Sean's asks. I would love to see actual Rust doc comments for this file: https://doc.rust-lang.org/rust-by-example/meta/doc.html. If you consider someone might be reading this code with no context about how these commands fit into the greater Bottlerocket projects, they may not be able to understand why the code operates in some ways.

deny.toml
bcressey: > We could introduce a lockfile to control when the CI will break. The benefit would be that we could see if introducing a new dependency introduces undue crate duplication downstream without randomly breaking CI. I advocated for disabling the multiple-versions check under the assumption that maintaining the lockfile wasn't worth the benefits.

If we need to do this, that's fine, but it should be in its own PR and not an unrelated one.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -385,7 +385,7 @@ mod test_etc_hosts_entries {
 /// character in user-facing identifiers. It stores the original form and makes it accessible
 /// through standard traits. Its purpose is to validate input for identifiers like container names
 /// that might be used to create files/directories.
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
```
bcressey, line 388: Why did this need `Ord` and `PartialOrd`?

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+        D: Deserializer<'de>,
+    {
+        let bootstrap_commands = HashMap::deserialize(deserializer)?;
+        Ok(Self { bootstrap_commands })
+    }
+}
+
+#[model(impl_default = true)]
+struct BootstrapCommand {
+    commands: Vec<ApiclientCommand>,
```
bcressey, line 36: nit: since we might expand this to other commands in the future:
```suggestion
    commands: Vec<AllowedCommand>,
```

(or `ValidCommand`, etc)

bcressey (commented):
Looks pretty close. If you rebase on develop, will that fix the build error?

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: We will be running BootstrapCommands in lexicographical order of their names. This allows sorting the keys in the Bootstrap Commands settings HashMap.

deny.toml
piyush-jena: Removed the commit from this PR. The PR that fixes this is PR #48. I will rebase to fix build errors.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
piyush-jena: Updated it to ValidCommand because it is similar to ValidBase64 in a different file.

bottlerocket-settings-models/modeled-types/src/shared.rs
bcressey: > This allows sorting the keys in the Bootstrap Commands settings HashMap.

I don't love the implicit coupling with an agent's implementation, which as a rule should be mediated by a config file rather than directly reusing these types. But I'll review https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/62 and see if it seems OK.

piyush-jena: This is the relevant line - https://github.com/piyush-jena/bottlerocket-core-kit/blob/d0fcf039f2d5d82755ba885b4f76cdf34de7b658/sources/api/bootstrap-commands/src/main.rs#L204

cbgbt: An escape hatch here is that any struct implemented via the `string_impls_for!` macro supports `ToString::to_string`.

On the implementation side, you could stringify these values prior to sorting them.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+impl ValidCommand {
+    pub fn get_args(&self) -> (String, Vec<String>) {
+        self.0
+            .clone()
+            .split_first()
+            .map_or(("".to_string(), Vec::new()), |(command, arguments)| {
+                (command.to_string(), arguments.to_vec())
+            })
+    }
+}
```
cbgbt: I have a few nits with this function:
1. The name is a bit misleading, since it really separates the command from the args
2. It unnecessarily clones data (both in the implementation as well as the interface).


For (1), you could implement this instead as two different functions `get_command()` and `get_args()`, or else I might rename this as `get_command_and_args()`.

For (2), the clone you have is unnecessary:

```suggestion
    pub fn get_args(&self) -> (String, Vec<String>) {
        self.0
            .split_first()
            .map_or(("".to_string(), Vec::new()), |(command, arguments)| {
                (command.to_string(), arguments.to_vec())
            })
    }
}
```

But the interface you've chosen (`(String, Vec<_>)`) will always require you to allocate your return values. I might recommend instead changing the interface to this:

```suggestion
    pub fn get_command_and_args(&self) -> (&str, &[String]) {
        self.0
            .split_first()
            .map(|(command, rest)| (command.as_str(), rest))
            .unwrap_or_else(Default::default)
    }
}
```

If possible, it's best to let the caller decide to allocate.

bottlerocket-settings-models/settings-models/src/lib.rs
```diff
@@ -27,6 +27,7 @@ pub use bottlerocket_string_impls_for as string_impls_for;
 pub use crate::boot::BootSettingsV1;
 pub use settings_extension_autoscaling::{self, AutoScalingSettingsV1};
 pub use settings_extension_aws::{self, AwsSettingsV1};
+pub use settings_extension_bootstrap_commands::{self, BootstrapCommandsSettingsV1, ValidCommand};
```
cbgbt: I wouldn't hoist `ValidCommand` to the public API here. `self` is added here so that callers can opt to reach in and grab additional models as needed.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
piyush-jena: I like your last implementation. `pub fn get_command_and_args(&self) -> (&str, &[String])`. It will help clean this up as well -> https://github.com/piyush-jena/bottlerocket-core-kit/blob/a4a72e2f1d6336c2669b4419bb1485f95cdfb824/sources/api/bootstrap-commands/src/main.rs#L113. args was a force of habit from C++ where argv contained the command as well as its first element.

cbgbt: For the linked code I would expect `String` to work without having to call `as_str` on everything, since `String` [has an impl for `AsRef<OsStr>`](https://doc.rust-lang.org/std/string/struct.String.html#impl-AsRef%3COsStr%3E-for-String).

Regardless, the `String` slice should work as well. Just call `iter()` on the result.

bottlerocket-settings-models/settings-models/src/lib.rs
cbgbt: Just to add to GitHub something we chatted about after I took a quick look at https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/62 -- there was some feedback about moving `ValidCommand` into the settings extension. The `bootstrap-command` agent in Bottlerocket must parse commands from its config file. Using the models here makes sense to validate the input, and we'd like to avoid having any dependencies between the agent and the extension directly.

This may make it wise to move this one struct back into the `modeled-types`, perhaps as something like `ApiclientCommand` so that it is more generic.

We'll discuss and see if we can come to some kind of consensus on the right API here.

bottlerocket-settings-models/settings-extensions/bootstrap-commands/src/lib.rs
piyush-jena: Updated with the last suggestion with a small change. `unwrap_or_default()` instead of `unwrap_or_else(Default::default)`

bottlerocket-settings-models/modeled-types/src/shared.rs
piyush-jena: We agreed on proceeding with Ord.

cbgbt (approved):
Nice work!

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -945,50 +945,124 @@ string_impls_for!(Lockdown, "Lockdown");
...
+#[cfg(test)]
+mod test_valid_apiclient_command {
+    use super::ApiclientCommand;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn valid_apiclient_command() {
+        assert!(ApiclientCommand::try_from(vec![
+            "apiclient".to_string(),
+            "motd=helloworld".to_string()
```
yeazelm, line 1002: nit: this isn't a valid command if you actually tried it though right? `apiclient motd=helloworld` will result in an error, it probably needs to be
```suggestion
            "get".to_string(),
            "motd=helloworld".to_string()
``` to actually be a valid command?

Comment (piyush-jena) [reply bottlerocket-settings-models/modeled-types/src/shared.rs:1002]: Updated. Thank you for spotting that.

File: bottlerocket-settings-models/modeled-types/src/shared.rs [hunk 23]
```diff
@@ -945,50 +945,125 @@ string_impls_for!(Lockdown, "Lockdown");
...
+    }
+}
+
+impl TryFrom<Vec<String>> for ApiclientCommand {
+    type Error = error::Error;
+
+    fn try_from(input: Vec<String>) -> std::result::Result<Self, error::Error> {
+        let first_word = input.first().map(String::as_str);
+        ensure!(
+            matches!(first_word, Some("apiclient")),
```
yeazelm, line 984: We probably should also confirm the second word isn't `exec` because we know that won't work for bootstrap commands since the host containers won't be up.

piyush-jena: I have updated it. But I am not sure if it should be here. On one hand, it **is** an API client command, but on the other hand, if allowed, it will result in a runtime failure. The correctness of a command is not really covered here. Should we let the failure happen during the configuration itself, or should we let it error out while it is running?

piyush-jena: As discussed, I will be adding details about this to bootstrap-commands README in the other PR. The intention will be to make users aware of the fact that valid apiclient commands that run in a session might fail here due to dependency on services that are yet to start until that point. The restriction itself won't be added here to keep it as general as possible.

piyush-jena: I have added the required information in bootstrap-commands binary in core-kit PR. It can be seen in the README.md as well.

bcressey (approved):

cbgbt (approved):

PR #49, opened 2024-08-07
feat: add nvidia shared GPU settings to k8s model 

monirul:
*Issue #, if available:*

*Description of changes:*
This PR introduces two new fields within the `NvidiaDevicePluginSettings` structure: `max_sharing_per_gpu` and `rename_shared_gpu`. These new fields are for managing the timeslicing capabilities of Nvidia GPUs.

Setting `max_sharing_per_gpu` to **1** deactivates time-slicing, effectively disabling GPU sharing. Conversely, a value greater than **1** activates time-slicing, resulting in the creation of the specified number of GPU replicas.

The `rename_shared_gpu` field dictates the naming convention for shared GPU resources. A `false` value retains the original GPU name, while `true` triggers a renaming process, appending `.shared` to the GPU identifier. For instance, enabling this feature would transform `nvidia.com/gpu` to `nvidia.com/gpu.shared`.

By contributing this pull request, I grant permission to use, modify, replicate, and distribute my submission, in accordance with your preferred terms.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice."

### DIFF
--- a/bottlerocket-settings-models/modeled-types/src/shared.rs
+++ b/bottlerocket-settings-models/modeled-types/src/shared.rs
@@ -1181,6 +1181,47 @@ mod test_positive_integer {
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+/// Input value that needs to be a positive integer, but should not be greater
+/// than an i32::MAX.
+#[derive(Clone, Debug, PartialEq, Scalar)]
+pub struct PositiveInteger {
+    inner: i32,
+}
+
+impl Validate for PositiveInteger {
+    fn validate<I: Into<i32>>(input: I) -> Result<PositiveInteger, ValidationError> {
+        let inner: i32 = input.into();
+        if inner < 1 {
+            Err(ValidationError::new(
+                "number must be great than or equal to 1",
+            ))
+        } else {
+            Ok(Self { inner })
+        }
+    }
+}
+
+#[cfg(test)]
+mod test_nonzero_integer {
+    use super::PositiveInteger;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn valid_positive_integer() {
+        assert!(PositiveInteger::try_from(1).is_ok());
+        assert!(PositiveInteger::try_from(i32::MAX).is_ok());
+        assert!(PositiveInteger::try_from(42).is_ok());
+    }
+
+    #[test]
+    fn invalid_positive_integer() {
+        assert!(PositiveInteger::try_from(i32::MIN).is_err());
+        assert!(PositiveInteger::try_from(-1).is_err());
+        assert!(PositiveInteger::try_from(0).is_err());
+    }
+}
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
 /// KernelCpuSetValue represents a string that contains a valid Kernel CpuSet Value from
 /// here: https://man7.org/linux/man-pages/man7/cpuset.7.html#FORMATS. This matches the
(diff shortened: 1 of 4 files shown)


bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -1181,6 +1181,47 @@ mod test_positive_integer {
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+/// Input value that needs to be a non-zero value, but should not be greater
+/// than an i32::MAX.
+#[derive(Clone, Debug, PartialEq, Scalar)]
+pub struct NonZeroInteger {
+    pub inner: i32,
+}
```
bcressey: You are also disallowing negative values. Maybe the type should be named `PositiveInteger`?

monirul: renamed.

cbgbt:
The code looks good to me! Do you mind rewriting the git commit message? I'll implement a linter for this, but the settings SDK tends to follow [the conventional commit style](https://www.conventionalcommits.org/en/v1.0.0/).

You could rewrite the message as:

```
feat: add nvidia shared GPU settings to k8s model
```

bottlerocket-settings-models/settings-models/src/kubernetes/mod.rs
```diff
@@ -24,6 +24,10 @@ use bottlerocket_modeled_types::{
     TopologyManagerScope, Url, ValidBase64, ValidLinuxHostname,
 };
 
+// Feature flags
+#[cfg(feature = "nvidia-device-plugin")]
```
cbgbt, line 28: We also need to set the variable to false in the absence of the feature.

cbgbt: Ah, you'll also need the variable to be `pub` to be accessible from the extension side.

monirul: I added that.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -1181,6 +1181,47 @@ mod test_positive_integer {
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+/// Input value that needs to be a positive integer, but should not be greater
+/// than an i32::MAX.
+#[derive(Clone, Debug, PartialEq, Scalar)]
+pub struct PositiveInteger {
+    pub inner: i32,
```
bcressey: Generally we should avoid direct access to `inner` fields like this, since it allows bypassing the intended type constraints.

You can just use `PositiveInteger::try_from` to construct instances, like you do in the unit test.

```suggestion
    inner: i32,
```

monirul: if I remove the pub, then following test will fail...
```
    --> bottlerocket-settings-models/modeled-types/src/kubernetes.rs:1493:65
     |
1493 |                     max_sharing_per_gpu: Some(PositiveInteger { inner: 10 }),
     |                                                                 ^^^^^^^^^ private field
```

cbgbt: The test could instead read:

```rust
    max_sharing_per_gpu: Some(PositiveInteger::try_from(10))
```

monirul: made the changes as suggested.

cbgbt:
Closing in favor of #63. Thanks!

PR #57, opened 2024-08-30
Add missing NVIDIA_DEVICE_PLUGIN_FEATURE_ENABLED

arnaldo2792:
*Issue #, if available:*
N / A

*Description of changes:*
A change to include the NVIDIA Settings API for kubernetes was merged, but this constant is needed by downstreams to prevent accidental builds of the API.

*Testing*:
I used my own Settings SDK, and confirmed that I had access to the constant

### DIFF
--- a/bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
+++ b/bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
@@ -21,6 +21,11 @@ use std::net::IpAddr;
 
 mod de;
 
+#[cfg(feature = "nvidia-device-plugin")]
+pub const NVIDIA_DEVICE_PLUGIN_FEATURE_ENABLED: bool = true;
+#[cfg(not(feature = "nvidia-device-plugin"))]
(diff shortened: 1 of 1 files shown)


vigh-m (approved):

cbgbt (approved):

sam-berning (approved):

arnaldo2792:
TL;DR: closing in favor of a top level `settings.kubelet-device-plugin` settings extension, rather than an `nvidia-device-plugin` cargo feature.

In the current implementation, we favored `nvidia-device-plugin` as a cargo feature so that downstream builders can extend their API definitions by enabling cargo features that extend the API model. However, we pivoted towards having a top level API along with its settings extension because we have use cases where the API shape conflicts with what the program expects. One example of this is `pluto` in the Bottlerocket Core Kit, it would have to keep enabling all the features added to the top level API for the deserialization to succeed, even though the program only cares about top level Kubernetes settings. 

To keep programs independent from whatever the downstream decide to consume in their API, we decided that it will be best to provide top level APIs instead of extending structs with cargo features.

PR #59, opened 2024-09-05
kubernetes: add max_parallel_image_pulls and serialize_image_pulls

ginglis13:
*Issue #, if available:*

https://github.com/bottlerocket-os/bottlerocket/issues/4058

*Description of changes:*

Add support for the `serializeImagePulls` and `maxParallelImagePulls` kubelet config settings: https://kubernetes.io/docs/concepts/containers/images/#maximum-parallel-image-pulls

once merged and release, I have changes staged for bottlerocket-core-kit to pass through the setting that I'll PR. Currently testing those. 


bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -98,6 +98,9 @@ pub struct KubernetesSettingsV1 {
     seccomp_default: bool,
     #[cfg(feature = "nvidia-device-plugin")]
     device_plugins: K8sDevicePluginsSettings,
+    // Generated in `k8s-1.27+` variants only
+    serialize_image_pulls: bool,
+    max_parallel_image_pulls: i32,
```
bcressey, line 103: Are these values not supported on older versions of `kubelet`?

If we add them like this, they'll still be in the settings API for older `*-k8s-*` builds.

ginglis13: Correct, this feature landed in k8s-1.27: https://kubernetes.io/docs/concepts/containers/images/#maximum-parallel-image-pulls.

> If we add them like this, they'll still be in the settings API for older *-k8s-* builds.

Also correct - my thought was that these settings would be ignored/unused in k8s < 1.27, and only included in a [`kubelet-config`](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/kubernetes-1.27/kubelet-config) in kubernetes-1.27+ packages. We'd call out this distinction on the bottlerocket.dev Settings Index

Open to suggestions on a better approach here, though

sam-berning: I think it makes sense to include the settings and document that they are ignored in older versions. Adding settings gated by a feature flag caused a few issues when we did it for `nvidia-device-plugin` settings, and in this case I don't think it makes sense to add a `kubernetes-1.27+` top-level setting.

cbgbt (approved):

PR #60, opened 2024-09-06
feat: drop nvidia-device-plugin feature

arnaldo2792:
*Issue #, if available:*
N / A

*Description of changes:*

Use a setings extension instead of a cargo feature for the kubelet device plugins API. See https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/57#issuecomment-2322174109 for details.

*Testing:*

As part of https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/132 and https://github.com/bottlerocket-os/bottlerocket/pull/4182

1. Instance joined the cluster:

```
NAME                                           STATUS   ROLES    AGE    VERSION
ip-XXXX.us-west-2.compute.internal   Ready    <none>   22s    v1.30.1-eks-e564799
```

2. Files were generated using the new values

```
bash-5.1# apiclient get settings.kubelet-device-plugin
{
  "settings": {
    "kubelet-device-plugin": {
      "nvidia": {
        "device-id-strategy": "index",
        "device-list-strategy": "volume-mounts",
        "pass-device-specs": true
      }
    }
  }
}

bash-5.1# cat /etc/systemd/system/nvidia-k8s-device-plugin.service.d/exec-start.conf
[Service]
ExecStart=
ExecStart=/usr/bin/nvidia-device-plugin --config-file=/etc/nvidia-k8s-device-plugin/settings.yaml
bash-5.1# cat /etc/nvidia-container-runtime/config.toml
### generated from the template file ###
accept-nvidia-visible-devices-as-volume-mounts = true
accept-nvidia-visible-devices-envvar-when-unprivileged = false

[nvidia-container-cli]
root = "/"
path = "/usr/bin/nvidia-container-cli"
environment = []
ldconfig = "@/sbin/ldconfig"
bash-5.1#
```

### DIFF
--- a/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
+++ b/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
@@ -1445,11 +1445,6 @@ mod test_hostname_override_source {
 }
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
-// Define the DevicePlugin struct
-#[model(impl_default = true)]
-pub struct K8sDevicePluginsSettings {
-    nvidia: NvidiaDevicePluginSettings,
-}
 
 /// NvidiaRuntimeSettings contains the container runtime settings for Nvidia gpu.
 #[model(impl_default = true)]
@@ -1476,26 +1471,6 @@ pub enum NvidiaDeviceListStrategy {
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    #[test]
-    fn test_serde_k8s_device_plugins() {
-        let test_json = r#"{"nvidia":{"pass-device-specs":true,"device-id-strategy":"index","device-list-strategy":"volume-mounts"}}"#;
-
-        let device_plugins: K8sDevicePluginsSettings = serde_json::from_str(test_json).unwrap();
-        assert_eq!(
-            device_plugins,
-            K8sDevicePluginsSettings {
-                nvidia: Some(NvidiaDevicePluginSettings {
-                    pass_device_specs: Some(true),
-                    device_id_strategy: Some(NvidiaDeviceIdStrategy::Index),
-                    device_list_strategy: Some(NvidiaDeviceListStrategy::VolumeMounts),
-                }),
-            }
-        );
-
-        let results = serde_json::to_string(&device_plugins).unwrap();
-        assert_eq!(results, test_json);
-    }
 
     #[test]
     fn test_serde_nvidia_device_plugins() {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ members = [
     "bottlerocket-settings-models/settings-extensions/host-containers",
     "bottlerocket-settings-models/settings-extensions/kernel",
     "bottlerocket-settings-models/settings-extensions/kubernetes",
+    "bottlerocket-settings-models/settings-extensions/kubelet-device-plugins",
     "bottlerocket-settings-models/settings-extensions/metrics",
     "bottlerocket-settings-models/settings-extensions/motd",
     "bottlerocket-settings-models/settings-extensions/network",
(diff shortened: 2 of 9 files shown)


bottlerocket-settings-models/settings-extensions/kubelet-device-plugin/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+//! Settings related to Amazon ECS
```
ginglis13: nit: this doc comment is not accurate

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1445,12 +1445,7 @@ mod test_hostname_override_source {
 }
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
-// Define the DevicePlugin struct
-#[model(impl_default = true)]
-pub struct K8sDevicePluginsSettings {
-    nvidia: NvidiaDevicePluginSettings,
-}
-
+//
```
ginglis13: nit: should this just be a blank line?

ginglis13 (approved):
LGTM with some nits. Quick question - I noticed in reading your testing in the description, why are we setting two `ExecStart=`, the first of which is empty?

https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/4b528d3ac4c9574befc053dc3f64c095234819c4/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin-exec-start-conf#L6-L7

arnaldo2792:
The first line is empty to reset the `ExecStart` commands. Otherwise, `systemd` will attempt to use the two lines and fail to start `nvidia-k8s-device-plugin.service` because the service type is `simple` which only accepts one single `ExecStart`.

bottlerocket-settings-models/settings-extensions/kubelet-device-plugin/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+use bottlerocket_model_derive::model;
+use bottlerocket_modeled_types::NvidiaDevicePluginSettings;
+use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
+use std::convert::Infallible;
+
+#[model(impl_default = true)]
+pub struct KubeletDevicePluginV1 {
```
bcressey: I'd prefer to expose this as `settings.kubelet-device-plugins` for consistency with our other plural settings clusters. So the type name should also be plural.

```suggestion
pub struct KubeletDevicePluginsV1 {
```

arnaldo2792:
Forced push includes:
- Fix nits
- Rename `KubeletDevicePluginV1` to `KubeletDevicePluginsV1`

bottlerocket-settings-models/settings-extensions/kubelet-device-plugin/src/main.rs
```diff
@@ -0,0 +1,18 @@
+use bottlerocket_settings_sdk::{BottlerocketSetting, NullMigratorExtensionBuilder};
+use settings_extension_kubelet_device_plugin::KubeletDevicePluginsV1;
```
bcressey:
```suggestion
use settings_extension_kubelet_device_plugins::KubeletDevicePluginsV1;
```

bcressey (approved):

arnaldo2792:
(forced push renames `kubelet-device-plugin` to `kubelet-device-plugins` and all the affected places)

cbgbt (approved):

bcressey (approved):

arnaldo2792:
I confirmed in a local build that `kubelet-device-plugins` is working as expected:

```
bash-5.1# apiclient get settings.kubelet-device-plugins
{
  "settings": {
    "kubelet-device-plugins": {
      "nvidia": {
        "device-id-strategy": "index",
        "device-list-strategy": "envvar",
        "pass-device-specs": true
      }
    }
  }
}
```

PR #62, opened 2024-10-01
feat: add nvidia time-slicing

KCSesh:
*Issue #, if available:*

*Description of changes:*
* Settings sdk changes for supporting settings.kubelet-device-plugins.nvidia.device-sharing-strategy and settings.kubelet-device-plugins.nvidia.time-slicing in bottlerocket.

*Testing:*
1. Instance joined the cluster

```
NAME                                           STATUS   ROLES    AGE   VERSION
ip-XXXX.us-west-2.compute.internal   Ready    <none>   15h   v1.29.5-eks-1109419
```

2. Model Default:

```
bash-5.1# apiclient get settings.kubelet-device-plugin
{
  "settings": {
    "kubelet-device-plugins": {
      "nvidia": {
        "device-id-strategy": "index",
        "device-list-strategy": "volume-mounts",
        "device-sharing-strategy": "none",
        "pass-device-specs": true,
      }
    }
  }
}
```

3. Model Updates:
```
bash-5.1#: apiclient set settings.kubelet-device-plugins.nvidia.device-sharing-strategy=time-slicing settings.kubelet-device-plugins.nvidia.time-slicing.replicas=2 settings.kubelet-device-plugins.nvidia.time-slicing.rename-by-default=true settings.kubelet-device-plugins.nvidia.time-slicing.fail-requests-greater-than-one=true
bash-5.1# apiclient get settings.kubelet-device-plugin
{
  "settings": {
    "kubelet-device-plugins": {
      "nvidia": {
        "device-id-strategy": "index",
        "device-list-strategy": "volume-mounts",
        "device-sharing-strategy": "none",
        "pass-device-specs": true,
        "time-slicing": {
          "fail-requests-greater-than-one": true,
          "rename-by-default": true,
          "replicas": 2
        }
      }
    }
  }
}
```

4. Bounded check:

```
bash-5.1# apiclient set settings.kubelet-device-plugins.nvidia.time-slicing.replicas=1
Failed to change settings: Failed PATCH request to '/settings/keypair?tx=apiclient-set-Ne2SavnsTF7Fweq0': Status 400 when PATCHing /settings/keypair?tx=apiclient-set-Ne2SavnsTF7Fweq0: Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error during deserialization: integer out of range, expected it to be between 2 and 2147483647 at line 1 column 107
bash-5.1#
```

5. Files generated:

```
bash-5.1# cat /etc/nvidia-k8s-device-plugin/settings.yaml
version: v1
flags:
  migStrategy: "none"
  failOnInitError: true
  plugin:
    passDeviceSpecs: true
    deviceListStrategy: volume-mounts
    deviceIDStrategy: index
sharing:
  timeSlicing:
    renameByDefault: true
    failRequestsGreaterThanOne: true
    resources:
    - name: "nvidia.com/gpu"
      replicas: 2
 ``` 
 
 ```
 nvidia-device-plugin[360007]:   "resources": {
nvidia-device-plugin[360007]:     "gpus": [
nvidia-device-plugin[360007]:       {
nvidia-device-plugin[360007]:         "pattern": "*",
nvidia-device-plugin[360007]:         "name": "nvidia.com/gpu"
nvidia-device-plugin[360007]:       }
nvidia-device-plugin[360007]:     ]
nvidia-device-plugin[360007]:   },
nvidia-device-plugin[360007]:   "sharing": {
nvidia-device-plugin[360007]:     "timeSlicing": {
nvidia-device-plugin[360007]:       "renameByDefault": true,
nvidia-device-plugin[360007]:       "failRequestsGreaterThanOne": true,
nvidia-device-plugin[360007]:       "resources": [
nvidia-device-plugin[360007]:         {
nvidia-device-plugin[360007]:           "name": "nvidia.com/gpu",
nvidia-device-plugin[360007]:           "rename": "nvidia.com/gpu.shared",
nvidia-device-plugin[360007]:           "devices": "all",
nvidia-device-plugin[360007]:           "replicas": 2
nvidia-device-plugin[360007]:         }
nvidia-device-plugin[360007]:       ]
nvidia-device-plugin[360007]:     }
nvidia-device-plugin[360007]:   }
nvidia-device-plugin[360007]: }
```


bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -20,6 +21,10 @@ use crate::SingleLineString;
 const IMAGE_GC_THRESHOLD_MAX: i32 = 100;
 const IMAGE_GC_THRESHOLD_MIN: i32 = 0;
 
+// Define the bounds for the `time-slicing.replicas` field
+const TIME_SLICING_REPLICAS_MIN: i32 = 2;
+const TIME_SLICING_REPLICAS_MAX: i32 = i32::MAX;
```
arnaldo2792, line 26: Does this limit correspond to the `timeSlicing` limit?

bottlerocket-settings-models/modeled-types/Cargo.toml
```diff
@@ -13,6 +13,7 @@ bottlerocket-scalar.workspace = true
 bottlerocket-scalar-derive.workspace = true
 bottlerocket-string-impls-for.workspace = true
 bottlerocket-model-derive.workspace = true
+bounded-integer = { workspace = true, features = ["types", "serde1"]}
```
arnaldo2792, line 16: Are you adding this dependency to validate the limit? If that's the case, I think a custom type would be best to prevent adding unnecessary dependencies. We already have custom types to validate upper and lower limits of certain values.

bottlerocket-settings-models/settings-extensions/kubelet-device-plugins/Cargo.toml
```diff
@@ -10,6 +10,7 @@ publish = false
 bottlerocket-modeled-types.workspace = true
 bottlerocket-model-derive.workspace = true
 bottlerocket-settings-sdk.workspace = true
+bounded-integer = { workspace = true, features = ["types"]}
```
bcressey: This could go in `dev-dependencies`, since it's only used for tests.

bottlerocket-settings-models/modeled-types/Cargo.toml
bcressey: > Are you adding this dependency to validate the limit? If that's the case, I think a custom type would be best to prevent adding unnecessary dependencies.

I like `bounded-integer` as a generic solution to the problem of restricting inputs to a certain number range. Otherwise we'll accumulate a lot of slightly different types over time; this one is like `NonNegativeInteger` but with an additional restriction.

cbgbt: Agreed! In general, I'm a big fan of dependencies that only add compile-time work and not runtime work -- using const generics is the right tool for the job here, but the implementation is more involved than you might think initially. It's great that someone implemented this already!

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
KCSesh: In the k8-device-plugin code they use an `int`, and verify it is over 2. 

In NVIDIA's technical blog it is marked as unlimited:
https://developer.nvidia.com/blog/improving-gpu-utilization-in-kubernetes/#virtualization_with_vgpu

KCSesh:
Moved `bounded-integer` in `settings-extensions` to dev-dependencies.

bottlerocket-settings-models/settings-extensions/kubelet-device-plugins/Cargo.toml
KCSesh: Updated, thanks for the suggestion.

cbgbt (approved):

bcressey (approved):

Cargo.toml
```diff
@@ -57,7 +57,7 @@ bottlerocket-template-helper = { path = "./bottlerocket-template-helper", versio
 
 # Settings Models
 bottlerocket-model-derive = { path = "./bottlerocket-settings-models/model-derive", version = "0.1" }
-bottlerocket-modeled-types = { path = "./bottlerocket-settings-models/modeled-types", version = "0.5" }
+bottlerocket-modeled-types = { path = "./bottlerocket-settings-models/modeled-types", version = "0.6" }
```
cbgbt, line 60: Sorry for the late comment, but we'll need to also bump the version of `settings-models`, which depends on this crate.

KCSesh: Sounds good, thanks for confirming.

KCSesh:
Updated `settings-models` to 0.6.0 as well.

cbgbt (approved):

PR #64, opened 2024-11-12
fix: align kubernetes cluster name validation with EKS

cartermckinnon:
*Issue #, if available:*

*Description of changes:*

This updates the `KubernetesClusterName` validation to align with EKS, per the `eks:CreateCluster` docs: https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html#API_CreateCluster_RequestSyntax

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #69, opened 2024-12-20
feat: Add kubernetes device ownership

vigh-m:
*Description of changes:*

* Create a new setting to hold the `device-ownership-from-security-context` boolean. This is to support the same setting in the containerd CRI plugin as described [here](https://kubernetes.io/blog/2021/11/09/non-root-containers-and-devices/)

Testing done:

* The setting shows up as expected in an AMI built using this `settings-sdk`
  ```
  # apiclient get settings.kubernetes.device-ownership-from-security-context
  {
    "settings": {
      "kubernetes": {
        "device-ownership-from-security-context": false
      }
    }
  }
  ```

* Changing the setting works as defined:
  ```
  [root@admin]# apiclient set settings.kubernetes.device-ownership-from-security-context="true"
  [root@admin]# apiclient get settings.kubernetes.device-ownership-from-security-context
  {
    "settings": {
      "kubernetes": {
        "device-ownership-from-security-context": true
      }
    }
  }
  [root@admin]# sheltie cat /etc/containerd/config.toml | grep device
  device_ownership_from_security_context = true
  ```

* The setting has the desired effect on device ownership
    * When `true`:
      ``` 
      # kubectl exec -it single-node-test -- bash
      ubuntu@single-node-test:/$ ls -lah /dev/ | grep neuron
      crw-------. 1 ubuntu 2000 244, 0 Dec 20 00:25 neuron0
      ```
    * When `false`:
      ``` 
      # kubectl exec -it single-node-test -- bash
      ubuntu@single-node-test:/$ ls -lah /dev/ | grep neuron
      crw-------. 1 root root 244, 0 Dec 19 20:03 neuron0
      ```

### DIFF
--- a/bottlerocket-settings-models/modeled-types/Cargo.toml
+++ b/bottlerocket-settings-models/modeled-types/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bottlerocket-modeled-types"
-version = "0.6.0"
+version = "0.7.0"
 authors = []
 license = "Apache-2.0 OR MIT"
 edition = "2021"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ bottlerocket-template-helper = { path = "./bottlerocket-template-helper", versio
 
 # Settings Models
 bottlerocket-model-derive = { path = "./bottlerocket-settings-models/model-derive", version = "0.1" }
-bottlerocket-modeled-types = { path = "./bottlerocket-settings-models/modeled-types", version = "0.6" }
+bottlerocket-modeled-types = { path = "./bottlerocket-settings-models/modeled-types", version = "0.7" }
 bottlerocket-scalar = { path = "./bottlerocket-settings-models/scalar", version = "0.1" }
 bottlerocket-scalar-derive = { path = "./bottlerocket-settings-models/scalar-derive", version = "0.1" }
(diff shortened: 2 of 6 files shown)


yeazelm:
The change looks good but nit about the commit message: I'd like a basic explanation about the change more than just the title to help folks piece together what it is intended for or why it was added since this is in a different repo from where the interesting parts will land.

yeazelm (approved):

vigh-m:
Ack. Updated the commit message with some more context.

ytsssun (approved):

bottlerocket-settings-models/modeled-types/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bottlerocket-modeled-types"
-version = "0.6.0"
+version = "0.7.0"
```
cbgbt, line 3: The `modeled-types` didn't actually change -- we want to bump the version for:
* The kubernetes settings extension
* The `settings-models` package.

That should be it.

vigh-m: Oh. I thought PR #64 changed the modeled-type and would need the version bump as a result. 

I've updated the settings extension.

cbgbt: You're right! My mistake.

cbgbt (approved):
Thanks!

PR #80, opened 2025-04-30
Add k8s setting container-log-max-workers and container-log-monitor-interval

ytsssun:
*Issue #, if available:*
https://github.com/bottlerocket-os/bottlerocket/issues/4470

*Description of changes:*
Add new Kubernetes settings for `containerLogMaxWorkers` and `containerLogMonitorInterval`.

Per the official doc:
> `ContainerLogMaxWorkers` specifies the maximum number of concurrent workers to spawn for performing the log rotate operations. Set this count to 1 for disabling the concurrent log rotation workflows. Default: 1
>  `containerLogMonitorInterval` specifies the duration at which the container logs are monitored for performing the log rotate operation. This defaults to 10 * time.Seconds. But can be customized to a smaller value based on the log generation rate and the size required to be rotated against. Default: 10s


piyush-jena (approved):

arnaldo2792 (approved):
It will be great if the commit message included a description of how the settings work or just some additional context

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -62,6 +62,8 @@ pub struct KubernetesSettingsV1 {
     kube_api_burst: i32,
     container_log_max_size: KubernetesQuantityValue,
     container_log_max_files: i32,
+    container_log_max_workers: i32,
+    container_log_monitor_internal: KubernetesDurationValue,
```
koooosh: Should be: `container_log_monitor_interval`, right?

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -165,6 +167,8 @@ mod test {
                 kube_api_burst: None,
                 container_log_max_size: None,
                 container_log_max_files: None,
+                container_log_max_workers: None,
+                container_log_monitor_internal: None,
```
koooosh: Should be: `container_log_monitor_interval`, right?

koooosh (changes requested):

KCSesh:
Missing a `bottlerocket-settings-models` version bump from `0.8.0` -> `0.9.0`. Right?

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
arnaldo2792: Can we also include `singleProcessOOMKill`?

arnaldo2792 (changes requested):

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
ytsssun: Addressed.

Discussed offline, we will add `singleProcessOOMKill` in separate PR.

koooosh (approved):
LGTM, assuming testing done in [the core-kit PR](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/487) still works

arnaldo2792 (approved):

KCSesh (approved):

cbgbt (approved):
We will need a migration when we take the related core-kit in the Bottlerocket repo.

ytsssun:
> We will need a migration when we take the related core-kit in the Bottlerocket repo.

Yeah I have drafted a migration implementation here - https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/80#pullrequestreview-2810827773 please ignore the bad snake case that @bcressey pointed out. I am going to test the migration next and then update the PR there.

PR #83, opened 2025-05-12
add: change K8s device list strategy from a single value to a list

sky1122:
*Description of changes:*
change K8s device list strategy from a single value to a list

*Testing done*

run `cargo test` in `/bottlerocket-settings-sdk/bottlerocket-settings-models`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ bottlerocket-template-helper = { path = "./bottlerocket-template-helper", versio
 
 # Settings Models
 bottlerocket-model-derive = { path = "./bottlerocket-settings-models/model-derive", version = "0.1" }
-bottlerocket-modeled-types = { path = "./bottlerocket-settings-models/modeled-types", version = "0.8" }
+bottlerocket-modeled-types = { path = "./bottlerocket-settings-models/modeled-types", version = "0.9" }
 bottlerocket-scalar = { path = "./bottlerocket-settings-models/scalar", version = "0.1" }
 bottlerocket-scalar-derive = { path = "./bottlerocket-settings-models/scalar-derive", version = "0.1" }
 bottlerocket-string-impls-for = { path = "./bottlerocket-settings-models/string-impls-for", version = "0.1" }
@@ -75,7 +75,7 @@ settings-extension-ecs = { path = "./bottlerocket-settings-models/settings-exten
 settings-extension-host-containers = { path = "./bottlerocket-settings-models/settings-extensions/host-containers", version = "0.1" }
 settings-extension-kernel = { path = "./bottlerocket-settings-models/settings-extensions/kernel", version = "0.1" }
 settings-extension-kubernetes = { path = "./bottlerocket-settings-models/settings-extensions/kubernetes", version = "0.3" }
-settings-extension-kubelet-device-plugins = { path = "./bottlerocket-settings-models/settings-extensions/kubelet-device-plugins", version = "0.2" }
+settings-extension-kubelet-device-plugins = { path = "./bottlerocket-settings-models/settings-extensions/kubelet-device-plugins", version = "0.3" }
 settings-extension-metrics = { path = "./bottlerocket-settings-models/settings-extensions/metrics", version = "0.1" }
 settings-extension-motd = { path = "./bottlerocket-settings-models/settings-extensions/motd", version = "0.1" }
 settings-extension-network = { path = "./bottlerocket-settings-models/settings-extensions/network", version = "0.1" }
--- a/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
+++ b/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
@@ -1481,10 +1481,40 @@ pub enum NvidiaDeviceIdStrategy {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
-#[serde(rename_all = "kebab-case")]
+#[serde(untagged)]
 pub enum NvidiaDeviceListStrategy {
+    Scalar(NvidiaDeviceListStrategyValues),
+    Vector(Vec<NvidiaDeviceListStrategyValues>),
+}
+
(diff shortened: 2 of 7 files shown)


sky1122:
forced push to change the last commit message from `bump` to `chore` in order to follow the Lint commit message

sky1122:
forced pushed to fix the lint, also run `make deny` on local to try to address the build error. but noting new changes add. will forced push again to resolve the new lint error

sky1122:
forced pushed to fix the max length of a subject and rebase on the build fix PR.

sky1122:
force pushed to change the version number

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1480,11 +1480,42 @@ pub enum NvidiaDeviceIdStrategy {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum NvidiaDeviceListStrategyValues {
     Envvar,
     VolumeMounts,
+    CdiCri,
+}
+
+impl NvidiaDeviceListStrategy {
+    pub fn iter<'a>(&'a self) -> Box<dyn Iterator<Item = &'a NvidiaDeviceListStrategyValues> + 'a> {
```
cbgbt: nit/rust tip: (no need to fix, just a good space to learn more about rust)

In general, we tend to avoid specifying types or lifetime boundaries that the compiler can deduce on its own when writing rust. The rules for lifetimes take a bit longer to get used to, but you can rely on a feature called [lifetime elision](https://doc.rust-lang.org/nomicon/lifetime-elision.html) here.

Basically, all data has a lifetime and the rust compiler keeps track of the lifetime of data that a reference points to. If the rust compiler can figure out what the lifetimes are, it won't require you to specify them.

Consider a common case like this getter method:


```rust
fn get_name(&self) -> &str;
```

This function only takes one pointer as input and returns one as output... we (and the compiler) can deduce that the output `&str` points to data that *MUST* share a lifetime with whatever `&self` points to (where else would we have gotten the pointer from?)

If Rust didn't have lifetime elision, we'd have to write that function signature like so:

```rust
// We're telling the compiler "there exists data with lifetime `'a` that
// &self references, and the resulting &str references some data with
// the same lifetime).
fn get_name<'a>(&'a self) -> &'a str;
```

Bringing this back to the code here, you also have a single pointer as input. Rust can make the same deduction, we just need to convince it to do so. Things are just a little unusual when you introduce `dyn Trait` in the return position, Rust requires that you specify the lifetime of the `dyn Trait` even though the compiler clearly could figure out that the lifetime must be the same as the input pointer!

```suggestion
    pub fn iter(&self) -> Box<dyn Iterator<Item = &NvidiaDeviceListStrategyValues> + '_> {
```

We're required to put *something* on the `dyn Trait`. But we can give `'_`, to force the compiler to elide this one too.

bottlerocket-settings-models/settings-extensions/kubelet-device-plugins/src/lib.rs
```diff
@@ -61,7 +61,7 @@ mod test {
 
     #[test]
     fn test_serde_kubelet_device_plugins() {
-        let test_json = r#"{"nvidia":{"pass-device-specs":true,"device-id-strategy":"index","device-list-strategy":"volume-mounts","device-sharing-strategy":"time-slicing","time-slicing":{"replicas":2,"rename-by-default":true,"fail-requests-greater-than-one":true},"device-partitioning-strategy":"mig","mig":{"profile":{"a100.40gb":"1g.5gb"}}}}"#;
+        let test_json = r#"{"nvidia":{"pass-device-specs":true,"device-id-strategy":"index","device-list-strategy":["volume-mounts"],"device-sharing-strategy":"time-slicing","time-slicing":{"replicas":2,"rename-by-default":true,"fail-requests-greater-than-one":true},"device-partitioning-strategy":"mig","mig":{"profile":{"a100.40gb":"1g.5gb"}}}}"#;
```
cbgbt: Could we change this test so that the vector has more than one item?

Likewise, could we re-add a test that uses a scalar?

cbgbt (commented):
Looks good, save for the test comment!

I'd also suggest squashing this into two commits:
* One with the feature addition, tests, and `cargo fmt`
* A second for the version updates.

If commits haven't already been merged, we tend to squash bugfixes within PRs into the same commit for the feature that introduced them.

sky1122:
force pushed to resolve the feedbacks

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
sky1122: Thanks for the sharing! learn a lot!!!

cbgbt (approved):
Looks good to me. It will be important for us to write a migration for this in https://github.com/bottlerocket-os/bottlerocket when we integrate this change.

sky1122:
force pushed to add unit test to make sure `serde` will keep the order when serialize/deserialized

sky1122:
force pushed to change the unit test location and fix rust 1.87 lint clippy lint.

ytsssun:
Could you provide a bit more context on this API shape change? 

What is the expected behavior when we have more than one item for NvidiaDeviceListStrategy?

arnaldo2792 (approved):

rpkelly (approved):
LGTM

Cargo.toml
ytsssun, line 1: Do you also need to bump the modeled-types version? 
https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/develop/bottlerocket-settings-models/modeled-types/Cargo.toml#L3

Also we seems to missed this in the past releases (README) - https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/develop/bottlerocket-settings-models/modeled-types/README.md?plain=1#L3

ytsssun (approved):

ytsssun (approved):

PR #86, opened 2025-05-23
chore: extend pr template with note on karpenter k8s settings

ginglis13:
*Description of changes:*

Add a note in PR template about extending Karpenter's settings model whenever Bottlerocket adds new settings under`settings.kubernetes`. Karpenter models Bottlerocket Kubernetes settings in their project [here](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/providers/amifamily/bootstrap/bottlerocketsettings.go) and any new setting is added to `settings.kubernetes` must also be added to the Karpenter model in order for users to be able to configure the setting via Karpenter userData. 

This PR template is the same as the bottlerocket-os org's template (https://github.com/bottlerocket-os/.github/blob/master/PULL_REQUEST_TEMPLATE) with checklist added to prompt PR authors to open a PR to github.com/aws/karpenter-provider-aws per-emptively with the setting.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,9 @@
+*Issue #, if available:*
+
+*Description of changes:*
+
+*Does this change affect `settings.kubernetes`?*
+- [ ] Yes, and I opened a PR to github.com/aws/karpenter-provider-aws to add the setting to Karpenter:
+- [x] No
+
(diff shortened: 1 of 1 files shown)


ytsssun:
Karpenter is extending the usage of `SettingRaw` (which is effectively map[string]interface{} parsed from the userdata toml) for also kubernetes settings.

I believe we can drop this once they merge their change upstream - https://github.com/aws/karpenter-provider-aws/pull/8254

PR #88, opened 2025-06-06
feat: k8s: add `memory-swap-behavior` setting

teskje:
This change adds a new `memory-swap-behavior` setting to control the k8s `memorySwap.swapBehavior` config. `memorySwap.swapBehavior` configures swap memory available to container workloads. By default containers are not allowed to use swap, so making this setting configurable is a requirement for using swap with Bottlerocket.

*Issue:* https://github.com/bottlerocket-os/bottlerocket/issues/4547

*Description of changes:*

This PR extends `KubernetesSettingsV1` with a new `memory-swap-behavior` setting, to control the `memorySwap.swapBehavior` kubelet config.

Testing the new setting with a custom Bottlerocket build is documented in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/541.

----------

### DIFF
--- a/bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
+++ b/bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
@@ -6,9 +6,10 @@ use bottlerocket_modeled_types::{
     KubernetesCloudProvider, KubernetesClusterDnsIp, KubernetesClusterName,
     KubernetesDurationValue, KubernetesEvictionKey, KubernetesHostnameOverrideSource,
     KubernetesLabelKey, KubernetesLabelValue, KubernetesMemoryManagerPolicy,
-    KubernetesMemoryReservation, KubernetesQuantityValue, KubernetesReservedResourceKey,
-    KubernetesTaintValue, KubernetesThresholdValue, NonNegativeInteger, SingleLineString,
-    TopologyManagerPolicy, TopologyManagerScope, Url, ValidBase64, ValidLinuxHostname,
+    KubernetesMemoryReservation, KubernetesMemorySwapBehavior, KubernetesQuantityValue,
+    KubernetesReservedResourceKey, KubernetesTaintValue, KubernetesThresholdValue,
+    NonNegativeInteger, SingleLineString, TopologyManagerPolicy, TopologyManagerScope, Url,
+    ValidBase64, ValidLinuxHostname,
 };
 use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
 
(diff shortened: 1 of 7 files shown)


bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -83,6 +90,7 @@ pub struct KubernetesSettingsV1 {
     memory_manager_reserved_memory: HashMap<Identifier, KubernetesMemoryReservation>,
     memory_manager_policy: KubernetesMemoryManagerPolicy,
     reserved_cpus: KernelCpuSetValue,
+    memory_swap: MemorySwapConfiguration,
```
bcressey: For consistency with the other Kubernetes settings in this model, I'd suggest flattening this into `memory_swap_behavior`, and setting it like:
```
[settings.kubernetes]
memory-swap-behavior = "NoSwap"
```

It might also be nice to define aliases, as `KubernetesMemoryManagerPolicy` does, so that we accept "no-swap", "limited-swap" in addition to the canonical camel-cased values.

teskje: Cool, will do! I was wondering about whether to flatten or not. Afaict, there is no other field where we had to make that decision yet, unless I'm missing one? There _is_ precedence for having nested options, but only on the form of mappings. But happy to defer to a more informed opinion!

teskje: Done.

teskje:
This is ready now. I've implemented @bcressey's suggestions above and tested the new setting with a custom Bottlerocket build (see https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/541).

One thing I'm unsure about are the crate version bumps. I cargo-culted them from https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/80, but then https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/81 introduced a new setting without bumping crate versions.

bcressey (approved):
@cbgbt can you advise on the versioning choices?

Otherwise LGTM.

ytsssun (approved):
Could you also bump the version in https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/develop/bottlerocket-settings-models/settings-models/Cargo.toml#L3 as the settings-models is updated too.

This would require a settings sdk release, which I can help with.

ytsssun:
The lint has a warning - could you address that by updating your commit message?

teskje:
Thanks, both done!

teskje:
Sorry, rustfmt was still complaining about imports being misordered. Fixed now as well.

cbgbt (approved):
Thanks!

PR #91, opened 2025-07-03
feat: add container-runtime-plugins settings model with SOCI parallel pull settings

### LINKED ISSUE #90: Add SOCI parallel_pull_unpack settings to settings models
**What I'd like**

* Add the settings for SOCI's new `parallel_pull_unpack` mode to settings models 
* These should exist under a new settings extension
* The `pull_mode` should be configurable, with the only option for now being `ParallelPullUnpack`
* The `decompress_streams` shouldn't be modeled as a similar approach to containerd's configuration in the bottlerocket-core-kit will be used based on system architecture: https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/0dfac2b2a4a5d25c2d06f4a775a31a7d85b0c47a/packages/containerd-1.7/containerd-1.7.spec#L73-L97

See the configuration docs on the soci-snapshotter repo: https://github.com/awslabs/soci-snapshotter/blob/main/docs/parallel-mode.md#configuration

ginglis13:
*Issue #, if available:*

Closes #90

*Description of changes:*

Add new settings extension for container runtime plugins with SOCI snapshotter configuration support. This includes modeling the parallel pull unpack settings for SOCI.

*Testing*

Added the ContainerRuntimePlugins settings extension to a dev variant:

```
[ssm-user@control]$ apiclient get settings.container-runtime-plugins
{
  "settings": {
    "container-runtime-plugins": {
      "soci-snapshotter": {
        "parallel-pull-unpack": {
          "concurrent-download-chunk-size": 8000000,
          "discard-unpacked-layers": true,
          "max-concurrent-downloads": 10,
          "max-concurrent-downloads-per-image": 5,
          "max-concurrent-unpacks": 5,
          "max-concurrent-unpacks-per-image": 3
        },
        "pull-mode": "parallel-pull-unpack"
      }
    }
  }
}
bash-5.1# apiclient get settings.container-runtime.snapshotter
{
  "settings": {
    "container-runtime": {
      "snapshotter": "soci"
    }
  }
}
```

Or as TOML:

```
[settings.container-runtime]
snapshotter="soci"

[settings.container-runtime-plugins.soci-snapshotter]
pull-mode = "parallel-pull-unpack"
[settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack]
concurrent-download-chunk-size = "8mb"
discard-unpacked-layers = true
max-concurrent-downloads = 10
max-concurrent-downloads-per-image = 5
max-concurrent-unpacks = 5
max-concurrent-unpacks-per-image = 3
```

Testing for concurrent-download-chunk-size:

```
bash-5.1# apiclient set settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.concurrent-download-chunk-size="1mb"
bash-5.1# cat /etc/soci-snapshotter/config.toml
...
[pull_modes.parallel_pull_unpack]
enable = true
max_concurrent_downloads = -1
max_concurrent_downloads_per_image = 3
concurrent_download_chunk_size = "1000000"
...

bash-5.1# journalctl -u soci-snapshotter | grep 10000
Jul 16 19:10:16 ip-192-168-62-23.us-west-2.compute.internal soci-snapshotter[24145]: {"concurrent_download_chunk_size":1000000,"discard_unpack_layers":false,"level":"info","max_concurrent_downloads":-1,"max_concurrent_downloads_per_image":3,"max_concurrent_unpacks":-1,"max_concurrent_unpacks_per_image":1,"msg":"Parallel image pull enabled","time":"2025-07-16T19:10:16.709907790Z"}

bash-5.1# apiclient set settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.concurrent-download-
bash-5.1# cat /etc/soci-snapshotter/config.toml
...
chunk-size=100
...
bash-5.1# journalctl -u soci-snapshotter | grep "max_conc"
Jul 16 19:13:46 ip-192-168-62-23.us-west-2.compute.internal soci-snapshotter[26113]: {"concurrent_download_chunk_size":100,"discard_unpack_layers":false,"level":"info","max_concurrent_downloads":-1,"max_concurrent_downloads_per_image":3,"max_concurrent_unpacks":-1,"max_concurrent_unpacks_per_image":1,"msg":"Parallel image pull enabled","time":"2025-07-16T19:13:46.018174687Z"}
```


ginglis13:
^ force push addresses the clippy lint to unrelated changes and a license clarification

ginglis13:
Going to split out clippy lints into a separate [PR](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/92) and rebase this once that's merged, as there are quite a few

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -1327,3 +1327,69 @@ mod test_kernel_cpu_set_value {
...
+/// The regex pattern is case-insensitive and allows optional whitespace.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct SizeString {
+    inner: String,
+}
+
+lazy_static! {
+    // https://github.com/awslabs/soci-snapshotter/blob/cb2b0e41acf3884de9a89a230deb4147edef72f5/config/parallel.go#L27
+    pub(crate) static ref SIZE_STRING: Regex =
+        Regex::new(r"(?i)^\s*(\d+(\.\d+)?)(\s*(gb|mb|kb|b))?\s*$").unwrap();
```
yeazelm: Why is the leading whitespace allowed? I see you got the regex from SOCI but it feels odd to allow leading whitespace in our datastore, even if technically allowed by SOCI? Is there any harm in making this a bit more strict since we own this type in our own code and API?

cbgbt: We could probably disallow trailing whitespace as well.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -1327,3 +1327,69 @@ mod test_kernel_cpu_set_value {
...
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// SizeString represents a string that contains a valid size specification with optional units.
+/// Valid formats include: "8b", "16kb", "64mb", "1gb", "1.5gb", "100", etc.
+/// The regex pattern is case-insensitive and allows optional whitespace.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct SizeString {
+    inner: String,
+}
```
cbgbt: May be nice to add a method on this to get the configured size as a [uom::si::information](https://crates.io/crates/uom) type. Could help to "re-use" this information elsewhere and make the implementation more generic.

Not a blocker though, for now we know exactly where this will be used and what the string form must be.

bottlerocket-settings-models/settings-models/Cargo.toml
```diff
@@ -29,6 +29,7 @@ settings-extension-bootstrap-containers.workspace = true
 settings-extension-cloudformation.workspace = true
 settings-extension-container-registry.workspace = true
 settings-extension-container-runtime.workspace = true
+settings-extension-container-runtime-plugins.workspace = true
```
cbgbt, line 32: We will need to bump the version of `settings-models` and `modeled-types` prior to releasing. Might as well add it in this PR.

bottlerocket-settings-models/modeled-types/src/shared.rs
ginglis13: I updated the regex to disallow any whitespace and modified the comment to indicate we've referenced the upstream regex but made modifications

ginglis13:
^ force push 

* adds the change for `settings.container-runtime` to include a snapshotter selection setting
* prepares for the next release

Originally I had intended to follow up with a different PR to include these commits once various work being tracked in https://github.com/bottlerocket-os/bottlerocket/issues/4578 had completed, but I've opted to instead keep everything in this PR and leave it in draft until all the integration work / testing for this has been approved.

bottlerocket-settings-models/modeled-types/src/shared.rs
```diff
@@ -1327,3 +1327,76 @@ mod test_kernel_cpu_set_value {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// SizeString represents a string that contains a valid size specification with optional units.
+/// Valid formats include: "8b", "16kb", "64mb", "1gb", "1.5gb", "100", etc.
+/// The regex pattern is case-insensitive and does not allow any whitespace.
```
bcressey: It's unfortunate that [SOCI uses nonstandard terminology](https://github.com/awslabs/soci-snapshotter/blob/8d8d1200b8eba2daf31a95ce8585a44a553d91cc/config/parallel.go#L29) - what they call "gb" is really "gib", etc.

It'd be worth looking at the `byte_unit` crate to parse the values in a standard way, and then turning the result into an i64 that we would write into the SOCI config (as a string value like it expects).

I would also really like to accept integers in the TOML settings for this, not just strings. The `OciDefaultsResourceLimitV1` struct has a custom deserializer that allows this, and that implementation could be copied and adapted to use `byte_unit` for its string parsing.

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+            ))
+        )
+    }
+
+    #[test]
+    fn test_serde_container_runtime_plugins() {
+        let test_json = json!({
+            "soci-snapshotter": {
+                "pull-mode": "parallel-pull-unpack",
+                "parallel-pull-unpack": {
```
bcressey, line 100: Seems like this got renamed from "parallel-pull-unpack" to just "parallel" in https://github.com/awslabs/soci-snapshotter/commit/4a78fc699fb41ca6401ac9b457b11bde4014b257.

ginglis13: The struct itself was renamed to `Parallel` but the actual TOML is still `parallel_pull_unpack`: https://github.com/awslabs/soci-snapshotter/commit/4a78fc699fb41ca6401ac9b457b11bde4014b257#diff-9c6f48c277a843e85b7c1a21f13bec4adc36b99204b8798bbd0a41ac120800e2R19-R24

we can still choose to model this as just `parallel` if that's preferred for its brevity

ginglis13:
^ force push uses the byte_unit crate for custom deserialization of `max-concurrent-downloads-per-image` and makes the underlying type `i64`.  Some manual testing details:

```
bash-5.1# apiclient set settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.concurrent-download-chunk-size="1mb"
bash-5.1# cat /etc/soci-snapshotter/config.toml
...
[pull_modes.parallel_pull_unpack]
enable = true
max_concurrent_downloads = -1
max_concurrent_downloads_per_image = 3
concurrent_download_chunk_size = "1000000"
...

bash-5.1# journalctl -u soci-snapshotter | grep 10000
Jul 16 19:10:16 ip-192-168-62-23.us-west-2.compute.internal soci-snapshotter[24145]: {"concurrent_download_chunk_size":1000000,"discard_unpack_layers":false,"level":"info","max_concurrent_downloads":-1,"max_concurrent_downloads_per_image":3,"max_concurrent_unpacks":-1,"max_concurrent_unpacks_per_image":1,"msg":"Parallel image pull enabled","time":"2025-07-16T19:10:16.709907790Z"}

bash-5.1# apiclient set settings.container-runtime-plugins.soci-snapshotter.parallel-pull-unpack.concurrent-download-
bash-5.1# cat /etc/soci-snapshotter/config.toml
...
chunk-size=100
...
bash-5.1# journalctl -u soci-snapshotter | grep "max_conc"
Jul 16 19:13:46 ip-192-168-62-23.us-west-2.compute.internal soci-snapshotter[26113]: {"concurrent_download_chunk_size":100,"discard_unpack_layers":false,"level":"info","max_concurrent_downloads":-1,"max_concurrent_downloads_per_image":3,"max_concurrent_unpacks":-1,"max_concurrent_unpacks_per_image":1,"msg":"Parallel image pull enabled","time":"2025-07-16T19:13:46.018174687Z"}
```

bottlerocket-settings-models/modeled-types/src/shared.rs
ginglis13: Latest force push refactors to using `byte_unit` rather than this regex and allows for integer or string (parsed as `byte_unit::Byte` or `i64`) to be provided as `concurrent-download-chunk-size`.

ginglis13:
^ force push removes `byte-unit.workspace = true` from `bottlerocket-settings-models/modeled-types/Cargo.toml`

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
```diff
@@ -0,0 +1,209 @@
...
+    match value {
+        StringOrIntOrNone::String(s) => {
+            let s = s.trim();
+
+            // Try parsing as a plain integer (bytes)
+            if let Ok(bytes) = s.parse::<i64>() {
+                if bytes < -1 {
+                    return Err(serde::de::Error::custom(format!(
+                        "Size cannot be less than -1: {s}"
+                    )));
```
cbgbt: If we're modifying our expectations of the incoming data, perhaps we should also do away with using `-1` to represent a lack of limits?

We could reserve a string for this, e.g. "unlimited"

ginglis13:
^ force push ensures `concurrent-download-chunk-size` can be `None` via `#[serde(default)]` and adds unit tests for this case

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
```diff
@@ -0,0 +1,209 @@
...
+pub fn deserialize_chunk_size<'de, D>(deserializer: D) -> Result<Option<i64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum StringOrIntOrNone {
+        String(String),
+        Int(i64),
+        None,
```
bcressey: I don't think we want to support explicit nulls in settings. The semantics around that would be a bit murky, like the value would be set, but to an "unset" value. Seems likely to trip something up.

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
```diff
@@ -0,0 +1,209 @@
...
+            } else {
+                s.to_string()
+            };
+
+            // Try parsing as a size string with units using byte_unit, ignoring case.
+            match Byte::parse_str(&normalized_s, true) {
+                Ok(byte) => {
+                    // Convert to i64, ensuring we don't overflow
+                    let bytes = byte.as_u64();
+                    Ok(Some(bytes as i64))
```
bcressey: This is an unchecked conversion that might overflow. Better to use `i64::try_from`.

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
```diff
@@ -0,0 +1,209 @@
...
+///
+/// Valid formats:
+/// - Integer values (e.g., "1024" or 1024)
+/// - Size strings with units (e.g., "64mb", "1.5gb", ".5gb")
+///
+/// Invalid formats:
+/// - Negative values other than -1
+/// - Strings with invalid formats
+/// - Strings with whitespace
+pub fn deserialize_chunk_size<'de, D>(deserializer: D) -> Result<Option<i64>, D::Error>
```
bcressey: We want the optionality to be applied to the struct field rather than to this value:
```suggestion
pub fn deserialize_chunk_size<'de, D>(deserializer: D) -> Result<i64, D::Error>
```

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+pub struct SociSnapshotterSettings {
+    pull_mode: SociSnapshotterPullMode,
+    parallel_pull_unpack: SociSnapshotterParallelPullUnpack,
+}
+
+#[model(impl_default = true)]
+pub struct SociSnapshotterParallelPullUnpack {
+    max_concurrent_downloads: i64,
+    max_concurrent_downloads_per_image: i64,
+    #[serde(deserialize_with = "deserialize_chunk_size", default)]
```
bcressey:
```suggestion
    #[serde(deserialize_with = "deserialize_chunk_size")]
```

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
cbgbt: In this same vein, given that we're taking some liberty with the data input, I wonder if it would be a good idea to support "unlimited" as an alternative to "-1".

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
ginglis13: I've used `-1` to represent a lack of limits to match the upstream configuration. IMO, we should keep it for that parity, but I'm open to also providing the option for specifying "unlimited" as an alias

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
ginglis13: Latest force push removes the explicit `None`.

ginglis13:
^ force pushes:
* Add a new wrapper around i64 `WrappedI64` for deserialization of `Option`
* handle u64->i64 conversion verbosely

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+    max_concurrent_downloads_per_image: i64,
+    concurrent_download_chunk_size: WrappedI64,
+    max_concurrent_unpacks: i64,
+    max_concurrent_unpacks_per_image: i64,
+    discard_unpacked_layers: bool,
+}
+
+/// WrappedI64 allows for handling Option<i64> with custom deserialization.
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, Eq, Ord, PartialOrd, PartialEq)]
+struct WrappedI64(#[serde(deserialize_with = "deserialize_chunk_size")] i64);
```
bcressey:
```suggestion
/// Handle Option<i64> with custom deserialization.
#[derive(Copy, Clone, Debug, Deserialize, Serialize, Eq, Ord, PartialOrd, PartialEq)]
struct ChunkSize(#[serde(deserialize_with = "deserialize_chunk_size")] i64);
```

bottlerocket-settings-models/settings-extensions/container-runtime-plugins/src/de.rs
ginglis13: > In this same vein, given that we're taking some liberty with the data input, I wonder if it would be a good idea to support "unlimited" as an alternative to "-1".

I'll add support for `"unlimited"` to the chunk size setting. There are some other settings here for which `-1` represents unlimited and is the default: `max-concurrent-downloads` and `max-concurrent-unpacks` - see https://github.com/awslabs/soci-snapshotter/blob/8d8d1200b8eba2daf31a95ce8585a44a553d91cc/config/defaults.go#L121-L132 However, these are i64 in the upstream soci-snapshotter config, as opposed to `concurrent-download-chunk-size` which is rendered as a string.

ginglis13:
^ force push supports `"unlimited"` to deserialize to `-1` for `concurrent-download-chunk-size`

ginglis13:
^ force push takes bug fix upstream into consideration (see https://github.com/awslabs/soci-snapshotter/pull/1651)
* add a new enum `ChunkSize` for modeling `concurrent-download-chunk-size`
* disallow negative values for this setting
* At render time in the soci config (see https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/569), if the chunk size is "unlimited", the set `concurrent_download_chunk_size=""` to indicate an unbounded value.

bcressey (approved):

cbgbt (approved):

yeazelm (approved):

bottlerocket-settings-models/CHANGELOG.md
```diff
@@ -9,7 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - See [unreleased changes here]
 
-[unreleased changes here]: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/compare/bottlerocket-settings-models-v0.11.0...HEAD
+[unreleased changes here]: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/compare/bottlerocket-settings-models-v0.12.0...HEAD
+
+## [0.12.0] - 2025-07-TODO
```
KCSesh: Just leaving a comment to ensure we fill out this todo.

ginglis13:
^ update README with release date, drop version bump from modeled types (no changes there after revisions to this PR)

PR #93, opened 2025-07-16
add a settings.kubernetes.static-pods-enabled setting

tzneal:
*Issue #, if available:*

*Description of changes:*

This will control the staticPodsPath for kubelet, allowing disabling it entirely to prevent static pod usage.


bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -93,14 +99,7 @@ pub struct KubernetesSettingsV1 {
     cluster_domain: DNSDomain,
     node_ip: IpAddr,
     pod_infra_container_image: SingleLineString,
```
tzneal, line 97: The comment on line 89 wasn't correct, so I re-organized these to fix it and also removed the `Generated in x` comments when its now in all versions due to version deprecation.

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -85,6 +85,12 @@ pub struct KubernetesSettingsV1 {
     memory_manager_policy: KubernetesMemoryManagerPolicy,
     reserved_cpus: KernelCpuSetValue,
     memory_swap_behavior: KubernetesMemorySwapBehavior,
+    hostname_override_source: KubernetesHostnameOverrideSource,
+    seccomp_default: bool,
+    device_ownership_from_security_context: bool,
+    // Generated in `k8s-1.32+` variants only
```
bcressey: nit: we can remove this comment also, and perhaps fold these settings into the rest of the list, plus the ones below where "we generate a value based on the runtime environment".

That's all in the realm of downstream builds at this point .

bcressey (approved):

ytsssun:
Do you mind also bumping the versions here - 

https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/develop/bottlerocket-settings-models/settings-extensions/kubernetes/Cargo.toml#L3

You can refer to this PR for it - https://github.com/bottlerocket-os/bottlerocket-settings-sdk/commit/bd85fcba8dbf244043333b60009333dedf1af0df#diff-d9c52eca38208221447a5ef8566e9da8b69f1c3694c010186cbf1e4c88bda45e

I think you only need to bump the version for settings-extension-kubernetes and bottlerocket-settings-models.

tzneal:
Yup, done

ytsssun (approved):

ytsssun (approved):

ginglis13 (approved):
LGTM

PR #94, opened 2025-07-18
feat: add `settings.{host,bootstrap}-containers.<name>.command` field

kasimeka:
*Issue #, if available:*
<https://github.com/bottlerocket-os/bottlerocket-core-kit/issues/593>

*Description of changes:*
adds a `command` field to the `host-containers` and `bootstrap-containers` model


bcressey (approved):

bottlerocket-settings-models/settings-extensions/host-containers/src/lib.rs
```diff
@@ -38,6 +38,7 @@ struct HostContainer {
     enabled: bool,
     superpowered: bool,
     user_data: ValidBase64,
+    command: Vec<String>,
```
bcressey: These might be safer as `Vec<SingleLineString>` - the `toml_encode` helper you're using over in the templates where the value is rendered makes this safe, but in theory the field could be used by some other template that might not expect newlines.

edit to add: oh, but in the implementation we eventually write to an environment file as `CTR_COMMAND={}` - so there's an implicit expectation that the value won't have newlines, which we should enforce at the type level here.

ytsssun:
This PR was created from the `develop` branch which is not recommended. I helped addressed the conflicts and resubmitted the PR here (with a feature branch) - https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/100 

Closing this PR in favor of the new one.

PR #105, opened 2025-10-27
Improve regex for NvidiaGpuModel

yeazelm:
*Description of changes:*

Some devices that support MIG have longer names than a single letter plus numbers such as the RTX PRO 6000. This change updates the regex to support these and future devices that have longer names.

### DIFF
--- a/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
+++ b/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
@@ -1581,7 +1581,7 @@ pub struct NvidiaGpuModel {
 }
 
 lazy_static! {
-    pub(crate) static ref NVIDIAGPU_NAME: Regex = Regex::new(r"^([a-z])(\d+)\.(\d+)gb$").unwrap();
+    pub(crate) static ref NVIDIAGPU_NAME: Regex = Regex::new(r"^([a-z0-9]+).(\d+)gb$").unwrap();
 }
 
 impl TryFrom<&str> for NvidiaGpuModel {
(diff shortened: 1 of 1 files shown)


bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1581,7 +1581,7 @@ pub struct NvidiaGpuModel {
 }
 
 lazy_static! {
-    pub(crate) static ref NVIDIAGPU_NAME: Regex = Regex::new(r"^([a-z])(\d+)\.(\d+)gb$").unwrap();
+    pub(crate) static ref NVIDIAGPU_NAME: Regex = Regex::new(r"^([a-z]+)(\d+)\.(\d+)gb$").unwrap();
```
jpculp:
```suggestion
    pub(crate) static ref NVIDIAGPU_NAME: Regex = Regex::new(r"^([a-z0-9]+)\.(\d+)gb$").unwrap();
```
Would it be safer to assume any combination of letters and numbers before the dot is valid? It looks like NVIDIA had a **a100x** at one point.

A stricter version could be `^([a-z]+)(\d+)0([a-z])?\.(\d+)gb$`

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1652,7 +1652,13 @@ mod test_nvidia_device_plugins {
 
     #[test]
     fn valid_gpu_model() {
-        for ok in &["a100.40gb", "a100.80gb", "h100.80gb", "h100.141gb"] {
+        for ok in &[
+            "a100.40gb",
+            "a100.80gb",
+            "h100.80gb",
+            "h100.141gb",
+            "rtxpro600.98gb",
```
jpculp:
```suggestion
            "rtxpro6000.98gb",
```

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
yeazelm: I think we can relax this more. I kept it to the known change I am aware of but we could relax this more to just alpha-numeric before the dot. Ultimately the migmanager will need to validate this.

jpculp (approved):

mgsharm (approved):

arnaldo2792 (approved):

piyush-jena (approved):

PR #107, opened 2025-12-31
feat: add MPS GPU sharing settings model

yeazelm:
*Issue #:* https://github.com/bottlerocket-os/bottlerocket/issues/4673

*Description of changes:*
Add NvidiaMpsSettings struct and Mps variant to NvidiaDeviceSharingStrategy for NVIDIA Multi-Process Service support. Includes validation to prevent MPS and MIG from being enabled simultaneously.

*Testing:*
The testing was documented in the related PR: https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/789

### DIFF
--- a/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
+++ b/bottlerocket-settings-models/modeled-types/src/kubernetes.rs
@@ -24,6 +24,10 @@ const IMAGE_GC_THRESHOLD_MIN: i32 = 0;
 // Define the bounds for the `time-slicing.replicas` field
 const TIME_SLICING_REPLICAS_MIN: i32 = 2;
 const TIME_SLICING_REPLICAS_MAX: i32 = i32::MAX;
+// Define the bounds for the `mps.replicas` field
+const MPS_REPLICAS_MIN: i32 = 2;
+// 48 is the current max supported on Volta and later cards
+const MPS_REPLICAS_MAX: i32 = 48;
 
 /// KubernetesName represents a string that contains a valid Kubernetes resource name.  It stores
 /// the original string and makes it accessible through standard traits.
@@ -1559,6 +1563,7 @@ pub struct NvidiaDevicePluginSettings {
     device_list_strategy: NvidiaDeviceListStrategy,
     device_sharing_strategy: NvidiaDeviceSharingStrategy,
     time_slicing: NvidiaTimeSlicingSettings,
+    mps: NvidiaMpsSettings,
     device_partitioning_strategy: NvidiaDevicePartitioningStrategy,
     mig: NvidiaMigSettings,
 }
@@ -1612,6 +1617,7 @@ impl IntoIterator for NvidiaDeviceListStrategy {
 pub enum NvidiaDeviceSharingStrategy {
     None,
     TimeSlicing,
+    Mps,
 }
 
 #[model(impl_default = true)]
@@ -1619,6 +1625,13 @@ pub struct NvidiaTimeSlicingSettings {
     replicas: BoundedI32<TIME_SLICING_REPLICAS_MIN, TIME_SLICING_REPLICAS_MAX>,
     rename_by_default: bool,
     fail_requests_greater_than_one: bool,
+}
+
+/// NvidiaMpsSettings contains the settings for NVIDIA Multi-Process Service (MPS) GPU sharing.
+#[model(impl_default = true)]
+pub struct NvidiaMpsSettings {
+    replicas: BoundedI32<MPS_REPLICAS_MIN, MPS_REPLICAS_MAX>,
+    rename_by_default: bool,
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
@@ -1762,6 +1775,7 @@ mod test_nvidia_device_plugins {
                 ),),
                 device_sharing_strategy: None,
                 time_slicing: None,
+                mps: None,
                 device_partitioning_strategy: None,
                 mig: None
             }
@@ -1785,6 +1799,7 @@ mod test_nvidia_device_plugins {
                 ),),
                 device_sharing_strategy: Some(NvidiaDeviceSharingStrategy::TimeSlicing),
                 time_slicing: None,
+                mps: None,
                 device_partitioning_strategy: None,
                 mig: None
             }
@@ -1816,6 +1831,7 @@ mod test_nvidia_device_plugins {
                 ),),
                 device_sharing_strategy: None,
                 time_slicing: None,
+                mps: None,
                 device_partitioning_strategy: Some(NvidiaDevicePartitioningStrategy::MIG),
                 mig: None
             }
@@ -1840,6 +1856,7 @@ mod test_nvidia_device_plugins {
                 ),),
                 device_sharing_strategy: None,
                 time_slicing: None,
+                mps: None,
                 device_partitioning_strategy: Some(NvidiaDevicePartitioningStrategy::MIG),
                 mig: Some(NvidiaMigSettings {
                     profile: Some(HashMap::from([(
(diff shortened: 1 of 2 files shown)


bottlerocket-settings-models/settings-extensions/kubelet-device-plugins/src/lib.rs
```diff
@@ -59,9 +81,63 @@ mod test {
...
+        let result = KubeletDevicePluginsV1::validate(settings, None);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_serde_kubelet_device_plugins_with_mps() {
+        let test_json = r#"{"nvidia":{"pass-device-specs":true,"device-id-strategy":"index","device-list-strategy":"volume-mounts","device-sharing-strategy":"mps","mps":{"replicas":4},"device-partitioning-strategy":"none"}}"#;
+
+        let device_plugins: KubeletDevicePluginsV1 = serde_json::from_str(test_json).unwrap();
+        assert_eq!(
```
piyush-jena, line 68: Can you add an assert for replicas? and an unit test for the default number of replicas = 2?

piyush-jena (approved):
LGTM apart from the nitpicky request for an additional unit test. Nice work coding the conflict between MIG and MPS.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -24,6 +24,9 @@ const IMAGE_GC_THRESHOLD_MIN: i32 = 0;
 // Define the bounds for the `time-slicing.replicas` field
 const TIME_SLICING_REPLICAS_MIN: i32 = 2;
 const TIME_SLICING_REPLICAS_MAX: i32 = i32::MAX;
+// Define the bounds for the `mps.replicas` field
+const MPS_REPLICAS_MIN: i32 = 2;
+const MPS_REPLICAS_MAX: i32 = i32::MAX;
```
bcressey: These are actual bounds? It's not possible to set only one replica, and the software is fine with 2 billion replicas?

bottlerocket-settings-models/settings-extensions/kubelet-device-plugins/src/lib.rs
```diff
@@ -33,8 +41,21 @@ impl SettingsModel for KubeletDevicePluginsV1 {
...
+                Some(NvidiaDeviceSharingStrategy::Mps)
+            );
+            let is_mig = matches!(
+                nvidia.device_partitioning_strategy,
+                Some(NvidiaDevicePartitioningStrategy::MIG)
+            );
+            if is_mps && is_mig {
+                return Err(KubeletDevicePluginsError::MpsMigConflict);
+            }
+        }
```
bcressey: Does this validation logic actually work? AFAIK it's the first time we're attempting this sort of cross-setting validation.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
yeazelm: It fails to start up if you specify under 2 for minimum. For max, I'm staying consistent with [time slicing](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/479b7bf511d194edc152b8974c335df82a36ddce/bottlerocket-settings-models/modeled-types/src/kubernetes.rs#L26).

The math described in [the doc discussing this implementation](https://docs.google.com/document/d/1H-ddA11laPQf_1olwXRjEDbzNihxprjPr74pZ4Vdf2M/edit?tab=t.0) asserts that the math is:
```
activeThreadPercentage = floor(100 / replicas)
pinnedMemoryLimit = floor(totalMemory / replicas)
```
The [code has a gate](https://github.com/NVIDIA/k8s-device-plugin/blob/9fc04c5d8bf70ab2c875540ea6dd52dd5d35daa5/cmd/mps-control-daemon/mps/device.go#L35) that should prevent you from setting over the limit and it looks to be dependent on hardware. The unit tests offer [some hints at current limitations from hardware](https://github.com/NVIDIA/k8s-device-plugin/blob/main/cmd/mps-control-daemon/mps/device_test.go#L25) at 16-48 max clients.

Setting to 50 currently fails on a g6.2xlarge since the hardware only supports 48. 
```
           "name": "nvidia.com/gpu",
           "rename": "nvidia.com/gpu.shared",
           "devices": "all",
           "replicas": 50
         }
       ]
     }
   },
   "imex": {}
 }
 I0112 17:15:15.683119 2713026 main.go:197] Retrieving MPS daemons.
 E0112 17:15:15.706465 2713026 main.go:84] error starting plugins: error getting daemons: invalid MPS configuration: invalid device maximum allowed replicas exceeded: 50 > 48
systemd[1]: nvidia-mps-control-daemon.service: Main process exited, code=exited, status=1/FAILURE
systemd[1]: nvidia-mps-control-daemon.service: Failed with result 'exit-code'.
systemd[1]: nvidia-mps-control-daemon.service: Scheduled restart job, restart counter is at 10.
```
This number could increase with future hardware so I'm not sure if we want to actually enforce it or just allow the NVIDIA software to enforce this number?

bottlerocket-settings-models/settings-extensions/kubelet-device-plugins/src/lib.rs
yeazelm: I can add the assert. We use the default 2 in the template if unset, so we are either setting to 2 if not set via the template or ensuring its within the bounds of 2-48. So there isn't really a way to test the default in this code.

bottlerocket-settings-models/settings-extensions/kubelet-device-plugins/src/lib.rs
cbgbt: That's correct. Validation right now only really holds at the serde model level. The next iteration of the extension design I had been evaluating removes some of these cross-settings validation concepts as well. It's not clear to me if it needlessly complicates the design. I'll take a closer look at these code changes, maybe they're a strong case for it.

cbgbt: As far as I can tell, we don't currently have the semantics to enforce something like this, even within a single model. This is because you can `PATCH` individual fields with the apiserver, which never loads the entire model to validate it - the validations occur at the field level. This would mean that to uphold this invariant, we either need to instrument changes to the apiserver or else design this structure to make the invalid combination impossible to express.

For the former, perhaps we could simply try to have the apiserver load the "full" settings model under a given keyspace (`settings.kubelet-device-plugins`) and let the serde deserializer run? Probably the most correct place to put this would be when a transaction is committed rather than when the settings themselves are sent - this would let someone stage an entire update to the model before applying it.

The backwards compatibility problem is tricky here though - what if folks are applying "partially correct" models intentionally right now? We could hedge this with a major version bump of the core-kit, perhaps.

cbgbt: I guess from the modeling perspective, is it possible for us to run MPS on a per-partition basis? Basically, can we do away with the feature incompatibility?

yeazelm: Right now the device plugin calls out that MPS and MIG should not be used together: "Sharing with MPS is currently not supported on devices with MIG enabled." [source](https://github.com/NVIDIA/k8s-device-plugin/blob/main/README.md?plain=1#L518). So we could just lean on the docs and refer upstream to the device plugin. In theory, if this ever changes when we update the device plugin it would also "just work" which is a bit more transparent rather than having to own the opinion here from a snapshot in time.

yeazelm: I moved the validation to the core kit in the template: https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/789. For the time being, I think this will suffice (it helps guide the user if they try with logs) but with or without my template change, the device plugin doesn't work with MIG and MPS enabled. This way its a bit more clear.

yeazelm:
^ Updated to remove the validation logic which isn't called like I expected it to be. Changed the max to 48 since that is what the device plugin currently enforces.

bcressey (approved):

PR #108, opened 2026-01-05
kubernetes: add fail-cgroup-v1 setting

ytsssun:
*Issue #, if available:*
https://github.com/bottlerocket-os/bottlerocket/issues/4725

*Description of changes:*
Add `fail-cgroup-v1` setting to support Kubernetes 1.35+ variants.

Kubernetes 1.35 enables `failCgroupV1=true` by default in kubelet, causing it to refuse starting on cgroup v1 nodes. While Bottlerocket defaults to cgroup v2, users who have opted into cgroup v1 via 
`settings.boot.init.systemd.unified_cgroup_hierarchy=0` will need a way to override this behavior.

This adds a new boolean setting:
```toml
[settings.kubernetes]
fail-cgroup-v1 = false
```

This allows users on cgroup v1 to continue running while they plan their migration to cgroup v2.


koooosh (commented):
We still need to bump the version [here](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/develop/bottlerocket-settings-models/settings-extensions/kubernetes/Cargo.toml), right?

ytsssun:
> We still need to bump the version [here](https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/develop/bottlerocket-settings-models/settings-extensions/kubernetes/Cargo.toml), right?

I was planning on doing that in the release PR, but I guess I can make them a single PR given the size of the diffs.

ytsssun:
Added changelog and bumped the model versions.

vigh-m (approved):

koooosh (approved):

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -100,6 +100,7 @@ pub struct KubernetesSettingsV1 {
     hostname_override: ValidLinuxHostname,
     ids_per_pod: KubernetesIdsPerPodValue,
     max_parallel_image_pulls: i32,
+    fail_cgroup_v1: bool,
```
KCSesh, line 104: Nit: The actual setting on kubelet is: `--fail-cgroupv1`, would we want `fail_cgroupv1` ; maybe as an alias? Not a blocker.

ytsssun: Synced offline, I think upstream were not being very consistent here as the corresponding field in kubelet config is `FailCgroupV1` instead of `FailCgroupv1`. 

But maybe to avoid the debate here, the best thing to do is have alias so that both formats are accepted.

ytsssun:
Force pushed to address comments from @KCSesh. Basically added alias to the setting.

KCSesh (approved):
LGTM

ytsssun:
Tested the alias - 

```
[root@admin]# sheltie
bash-5.1# apiclient get settings.kubernetes.fail
{}
bash-5.1# apiclient set settings.kubernetes.fail-cgroupv1=false
bash-5.1# apiclient get settings.kubernetes.fail
{
  "settings": {
    "kubernetes": {
      "fail-cgroup-v1": false
    }
  }
}
bash-5.1# apiclient set settings.kubernetes.fail-cgroup-v1=false
bash-5.1# apiclient set settings.kubernetes.fail-cgroup-v1=true
bash-5.1# apiclient get settings.kubernetes.fail
{
  "settings": {
    "kubernetes": {
      "fail-cgroup-v1": true
    }
  }
}
bash-5.1#
```

PR #109, opened 2026-01-07
kubernetes: fix fail-cgroup-v1 serialization

ytsssun:
*Issue #, if available:*
https://github.com/bottlerocket-os/bottlerocket/issues/4725

*Description of changes:*
Fix serialization of fail-cgroup-v1 setting added in #108.

The field was missing `skip_serializing_if = "Option::is_none"` which caused it to serialize as null instead of being omitted when not set. This broke the `test_default_kubernetes_settings_empty` test in core-kit that rely on default settings serializing to an empty object.


ytsssun:
Closing this PRas it was created based off a wrong branch.

PR #111, opened 2026-01-14
Revert fail-cgroup-v1 setting

ytsssun:
*Issue #, if available:*
N/A

*Description of changes:*
We introduced the `fail-cgroup-v1` setting in https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/108. But we ended up not taking it for k8s 1.35 variants. This PR is to revert the settings. Instead we decided to set the `FailCgroupV1` field in kubelet config to `false`.

We have verified that cgroupv1 works with k8s 1.35 variant with the core-kit https://github.com/bottlerocket-os/bottlerocket-core-kit/releases/tag/v12.2.0.

See the test details in https://github.com/bottlerocket-os/bottlerocket/pull/4735


bottlerocket-settings-models/CHANGELOG.md
```diff
@@ -9,7 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-[unreleased changes here]: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/compare/bottlerocket-settings-models-v0.19.0...HEAD
+[unreleased changes here]: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/compare/bottlerocket-settings-models-v0.20.0...HEAD
+
+## [0.20.0] - 2026-01-14
+
+## Model Changes
+
+### Removed
+
+- Removed `fail-cgroup-v1` kubernetes setting ([#111])
```
koooosh: minor nit: I prefer "revert" instead of "remove" here since "remove" gives me the impression that this setting was at one point included in a Bottlerocket ami, which is not the case.

koooosh (approved):

bottlerocket-settings-models/CHANGELOG.md
ytsssun: Good point, just updated!

koooosh (approved):

yeazelm (approved):

Issue #112, opened 2026-01-20
Add serialization tests for settings models to catch missing `skip_serializing_if`


ytsssun:
### Background

PR #110 attempted to fix a bug where fail_cgroup_v1 was serializing as "fail_cgroup_v1": null instead of being omitted when None. The field was missing `skip_serializing_if = "Option::is_none"`.

This fix attempt revealed a gap in our test coverage: we only verify deserialization (JSON → struct), not serialization (struct → JSON). The bug would have been caught immediately with proper serialization tests.

### Gap

Settings models lack tests that verify serialization output. Specifically:

1. Default serialization - A default-constructed settings struct should serialize to {}, not include null values for optional fields
2. Round-trip consistency - Deserialize → serialize should produce equivalent JSON

### Proposed Solution

Add serialization tests for settings models, below is a very simplified version, we could take it from there and make it more comprehensive.

```rust
#[test]
fn test_default_serialization() {
    let settings = MySettings::default();
    let json = serde_json::to_value(&settings).unwrap();
    assert_eq!(json, serde_json::json!({}));
}
```

This pattern should be applied to all settings models in the settings SDK to prevent similar regressions.

### References

- https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/110

PR #114, opened 2026-02-02
feat: make image verifer plugins extensible

bcressey:
**Issue #, if available:**
Related: https://github.com/bottlerocket-os/bottlerocket/issues/4684

**Description of changes:**
Replace the hard-coded "notation" plugin with an extensible map. This allows additional plugins to be added by downstream builds, or at runtime by copying plugins to a host path.

Plugins can store a corresponding trust policy in the settings API, which must be a base64-encoded JSON blob. Although this is optional, registering the trust policy with the OS ensures that it is measured into the TPM2 device, making it visible for remote attestation. The JSON format ensures that the policy can be canonicalized at runtime, rather than requiring it to be provied in canonical form.

**Testing done:**

Verified that `enabled` can't be the name of a plugin:
```
bash-5.3# apiclient apply <<EOF
[settings.image-verifier-plugins.enabled]
trustpolicy = "eyJ0cnVzdGVkRGlnZXN0cyI6W119"
EOF
Failed to apply settings: Failed to PATCH settings from '-' to '/settings?tx=apiclient-apply-oUh6I9Nnfo7wmvYy': Status 400 when PATCHing /settings?tx=apiclient-apply-oUh6I9Nnfo7wmvYy: Json deserialize error: invalid type: map, expected a boolean at line 1 column 37
```

Verified that I can configure two other policies:
```
bash-5.3# apiclient apply <<EOF
[settings.image-verifier-plugins]
enabled = true

[settings.image-verifier-plugins.digestion]
trustpolicy = "..."

[settings.image-verifier-plugins.notation]
trustpolicy = "..."
EOF

bash-5.3# apiclient get settings.image-verifier-plugins
{
  "settings": {
    "image-verifier-plugins": {
      "digestion": {
        "trustpolicy": "..."
      },
      "enabled": true,
      "notation": {
        "trustpolicy": "..."
      }
    }
  }
```

**Terms of contribution:**

### DIFF
--- a/bottlerocket-settings-models/settings-extensions/image-verifier-plugins/src/lib.rs
+++ b/bottlerocket-settings-models/settings-extensions/image-verifier-plugins/src/lib.rs
@@ -1,21 +1,26 @@
 //! Settings related to Image Verifier Plugins
 
 use bottlerocket_model_derive::model;
-use bottlerocket_modeled_types::ValidBase64;
+use bottlerocket_modeled_types::{Identifier, ValidBase64Json};
 use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::convert::Infallible;
 
-#[model(impl_default = true)]
+/// Settings for image verifier plugins
+#[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
 pub struct ImageVerifierPluginsSettingsV1 {
-    enabled: bool,
-    notation: NotationSettings,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub enabled: Option<bool>,
+    #[serde(flatten)]
+    pub plugins: HashMap<Identifier, VerifierPluginConfig>,
 }
 
+/// Configuration for a verifier plugin
 #[model(impl_default = true)]
-pub struct NotationSettings {
-    /// Base64 encoded trustpolicy.json
-    /// https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trust-store
-    trustpolicy: ValidBase64,
+pub struct VerifierPluginConfig {
+    #[serde(alias = "trust-policy", skip_serializing_if = "Option::is_none")]
+    trustpolicy: ValidBase64Json,
 }
 
 type Result<T> = std::result::Result<T, Infallible>;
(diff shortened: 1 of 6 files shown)


ginglis13 (approved):

bcressey:
I've also verified that using this settings SDK, with no other changes to kits, still works for existing variants like `aws-ecs-3` that have the earlier, Notation-only logic.

bottlerocket-settings-models/settings-extensions/image-verifier-plugins/src/lib.rs
```diff
@@ -1,21 +1,26 @@
...
 }
 
+/// Configuration for a verifier plugin
 #[model(impl_default = true)]
-pub struct NotationSettings {
-    /// Base64 encoded trustpolicy.json
-    /// https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trust-store
-    trustpolicy: ValidBase64,
+pub struct VerifierPluginConfig {
+    #[serde(alias = "trust-policy")]
```
ytsssun: Adding `alias` would overwrite the `impl_default`. So this needs to be paired with
```suggestion
    #[serde(alias = "trust-policy", skip_serializing_if = "Option::is_none")]
```
Basically if not set in userdata, the serialized output json would not skip `trust-policy` which is not the desired default for `impl_default`. 

See the GH issue for more context - https://github.com/bottlerocket-os/bottlerocket-settings-sdk/issues/112

For kubernetes settings, I hit [this failure](https://github.com/bottlerocket-os/bottlerocket-core-kit/actions/runs/20778407660/job/59669724442) before (in core-kit). But I don't think we have any unit tests capture that for `ImageVerifierPluginsSettingsV1`.
```
[1](https://github.com/bottlerocket-os/bottlerocket-core-kit/actions/runs/20778407660/job/59669724442#step:7:1522)
thread 'api::test::test_default_kubernetes_settings_empty' (28454) panicked at api/pluto/src/api.rs:188:9:
assertion `left == right` failed
  left: Object {"fail-cgroup-v1": Null}
 right: Object {}
```

bcressey:
Add fix from @ytsssun and test cases to confirm.

ytsssun:
Approved assuming CI passed.

KCSesh (approved):

ytsssun (approved):

PR #117, opened 2026-03-03
Add k8s settings: prefer-closest-numa-nodes and max-allowable-numa-nodes

piyush-jena:
Issue # https://github.com/bottlerocket-os/bottlerocket/issues/4750

Related to: 
- https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/848
- https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/117
- https://github.com/bottlerocket-os/bottlerocket/pull/4778

*Description of changes:*
Add 2 topology manager policy options:
1. `max-allowable-numa-nodes` - GA k8s-1.35+
2. `prefer-closest-numa-nodes` - GA k8s-1.32+

Testing details available in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/848


mgsharm:
Can you add sign-off in the commit.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1555,6 +1555,66 @@ mod test_kubernetes_ids_per_pod_value {
...
+            KubernetesMaxAllowableNumaNodesValue::try_from(*err).unwrap_err();
+        }
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesTopologyManagerPolicyOptions represents the topology manager policy options
+/// for the kubelet. These are rendered as `topologyManagerPolicyOptions` in KubeletConfiguration.
+/// Upstream source: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/policy_options.go
```
ytsssun: Nit: can we use a permanent link here?

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
```diff
@@ -71,6 +72,7 @@ pub struct KubernetesSettingsV1 {
     cpu_manager_policy_options: Vec<KubernetesCPUManagerPolicyOption>,
     topology_manager_scope: TopologyManagerScope,
     topology_manager_policy: TopologyManagerPolicy,
+    topology_manager_policy_options: KubernetesTopologyManagerPolicyOptions,
```
mgsharm, line 75: Should this be TopologyManagerPolicyOptions? The sibling types TopologyManagerScope and TopologyManagerPolicy don't have the Kubernetes prefix, and the parent struct already establishes the Kubernetes context.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1555,6 +1555,66 @@ mod test_kubernetes_ids_per_pod_value {
...
+}
+
+impl TryFrom<i32> for KubernetesMaxAllowableNumaNodesValue {
+    type Error = error::Error;
+
+    fn try_from(input: i32) -> Result<Self, Self::Error> {
+        ensure!(
+            input >= 8,
+            error::InvalidKubernetesMaxAllowableNumaNodesValueSnafu { input }
+        );
```
ytsssun, line 1574: Not blocking, but I wonder if we should loosen the validation here and let kubelet do the validation instead. I don't see this ">= 8" requirement explicitly mentioned in their doc. For that reason, it does not feel like something they would commit to. And if they happened to change their validation logic here, the extra validation we are doing here will become stale.

bottlerocket-settings-models/CHANGELOG.md
```diff
@@ -9,7 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+## [0.22.0] - 2026-03-07
+
+## Model Changes
+
+### Added
+
+- Added `topology-manager-policy-options` kubernetes setting with `prefer-closest-numa-nodes` (bool) and `max-allowable-numa-nodes` (constrained integer >= 8) sub-fields
+- Added `KubernetesMaxAllowableNumaNodesValue` constrained integer type
+- Added `KubernetesTopologyManagerPolicyOptions` struct
+
```
mgsharm, line 23: nit: We can remove these lines as these are implementation details. Users of changelog only care about the settings.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1555,6 +1555,66 @@ mod test_kubernetes_ids_per_pod_value {
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
+/// KubernetesMaxAllowableNumaNodesValue represents a valid value for the topology manager
+/// `max-allowable-numa-nodes` policy option. Must be >= 8 (the default). Upstream validation:
+/// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/policy_options.go
```
arnaldo2792: This has to be a permalink, not a link to the main branch.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
arnaldo2792: This is the hard-coded validation:

https://github.com/kubernetes/kubernetes/blob/f999104ffe0107a33058d022ac6d14279e31390e/pkg/kubelet/cm/topologymanager/topology_manager.go#L41

Historically, Bottlerocket's API doesn't allow set "invalid" values that will prevent the kubelet from starting, the API server hard fails if the settings are invalid for the kubelet.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
```diff
@@ -1555,6 +1555,66 @@ mod test_kubernetes_ids_per_pod_value {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(try_from = "i32", into = "i32")]
+pub struct KubernetesMaxAllowableNumaNodesValue {
+    inner: i32,
```
arnaldo2792: Shouldn't this be `u32` instead?

bottlerocket-settings-models/settings-extensions/kubernetes/src/lib.rs
piyush-jena: The more closer equivalent setting is the `cpu_manager_policy_options` of type `Vec<KubernetesCPUManagerPolicyOption>`. Maybe that's why AI generated it this way but your comment definitely applies to `KubernetesMaxAllowableNumaNodesValue`.

piyush-jena: After going through other settings it seems like Kubernetes prefix is used in other types as well and seems to be the appropriate one here as described above. However, there is no good resource that tells us when this should be used. But I think setting types seem to have the prefix and the underlying settings don't in most cases.

bottlerocket-settings-models/CHANGELOG.md
piyush-jena: True. Updated!

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
piyush-jena: Updated to permalink.

bottlerocket-settings-models/modeled-types/src/kubernetes.rs
piyush-jena: In the upstream they have used int variable here - https://github.com/kubernetes/kubernetes/blob/473b7635de1d1af0fe2663922e1ca042c99cd1fa/pkg/kubelet/cm/topologymanager/policy_options.go#L61 even though unsigned exists in go which is why (A)I used i32 but I updated it to unsigned int because that's more correct.

bottlerocket-settings-models/CHANGELOG.md
```diff
@@ -9,7 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-[unreleased changes here]: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/compare/bottlerocket-settings-models-v0.21.0...HEAD
+[unreleased changes here]: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/compare/bottlerocket-settings-models-v0.22.0...HEAD
+
+## [0.22.0] - 2026-03-30
+
+## Model Changes
+
+### Added
+
+- Added support for `topology-manager-policy-options` kubernetes settings `prefer-closest-numa-nodes` and `max-allowable-numa-nodes` ([#117])
```
arnaldo2792: I don't understand this line. Shouldn't it be:

```suggestion
Add support for  `prefer-closest-numa-nodes` and `max-allowable-numa-nodes` in Kubernetes' Topology Manager Policy options
```

ytsssun (approved):

mgsharm (approved):

piyush-jena:
force push addresses @arnaldo2792 's suggestion.

arnaldo2792 (approved):

PR #140, opened 2026-07-02
feat: add settings.measurement model for PCR 8 exclusion control

ginglis13:
*Issue #, if available:*

Related: https://github.com/bottlerocket-os/bottlerocket/issues/4872

*Description of changes:*

Add a new settings extension for `settings.measurement` that allows customers to configure which settings are excluded from PCR 8 measurement. Includes validation to reject dangerous patterns (bare "settings", empty strings, wildcards, and self-references).

Values are accepted with or without a "settings." prefix for convenience — both "host-containers" and "settings.host-containers" are treated equivalently.

### DIFF
--- a/bottlerocket-settings-models/settings-extensions/measurement/src/lib.rs
+++ b/bottlerocket-settings-models/settings-extensions/measurement/src/lib.rs
@@ -1,0 +1,284 @@
+//! Settings for controlling what is excluded from TPM PCR measurements.
+//!
+//! The `settings.measurement` namespace lets customers configure which settings
+//! are excluded from PCR 8 measurement by rottweiler.
+//!
+//! Each entry in `excluded-settings` names a settings node. Specifying a node
+//! excludes it and the entire tree beneath it — for example, `"host-containers"`
+//! excludes all settings under `settings.host-containers`. Wildcards are not
+//! supported; specify the node directly.
+use bottlerocket_model_derive::model;
+use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
+use snafu::{ensure, Snafu};
+
+/// Patterns that must never be allowed as exclusions because they would
+/// undermine the integrity of the measurement system.
+const DISALLOWED_PATTERNS: &[&str] = &["settings", ""];
+
+#[derive(Debug, Snafu, PartialEq)]
+pub enum MeasurementValidationError {
+    #[snafu(display("Disallowed measurement exclusion pattern: {:?}", pattern))]
+    DisallowedPattern { pattern: String },
+
+    #[snafu(display("Self-referencing measurement exclusion not allowed: {:?}", pattern))]
+    SelfReference { pattern: String },
+
+    #[snafu(display("Wildcard not allowed in measurement exclusion: {:?} — specify the top-level setting node to exclude its entire tree", pattern))]
+    ContainsWildcard { pattern: String },
+}
+
+/// Normalize an exclusion value by stripping a leading "settings." prefix if present.
+/// Users may provide either "host-containers" or "settings.host-containers".
+fn normalize_exclusion(value: &str) -> &str {
+    value.strip_prefix("settings.").unwrap_or(value)
+}
+
+/// Check whether a normalized exclusion is a self-reference to the measurement namespace.
+fn is_self_reference(normalized: &str) -> bool {
+    normalized == "measurement" || normalized.starts_with("measurement.")
+}
+
+/// Validate that a single exclusion pattern is allowed.
+fn validate_exclusion(value: &str) -> std::result::Result<(), MeasurementValidationError> {
+    let trimmed = value.trim();
+    let normalized = normalize_exclusion(trimmed);
+    ensure!(
+        !DISALLOWED_PATTERNS.contains(&normalized),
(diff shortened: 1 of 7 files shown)


bottlerocket-settings-models/settings-extensions/measurement/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+        let input = vec![String::from("*"), String::from("host-containers")];
+        let result = filter_excluded_settings(&input);
+        assert_eq!(result, vec!["host-containers"]);
+    }
+
+    #[test]
+    fn test_filter_rejects_self_reference() {
+        let input = vec![
+            String::from("measurement"),
+            String::from("measurement.excluded-settings"),
```
piyush-jena: still building an understanding of the code but it looks like we are silently ignoring bad values.

bottlerocket-settings-models/settings-extensions/measurement/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+    value.strip_prefix("settings.").unwrap_or(value)
+}
+
+/// Filter out disallowed exclusion patterns, logging a warning for each one.
+/// Values are normalized (leading "settings." stripped) before checking.
+pub fn filter_excluded_settings(values: &[String]) -> Vec<String> {
+    values
+        .iter()
+        .filter_map(|v| {
+            let normalized = normalize_exclusion(v.trim());
```
piyush-jena: We have `v.trim()` 3 times in this block. Do you think it would be good to add `let excluded_setting = v.trim();` and use that value in rest of the places?

bottlerocket-settings-models/settings-extensions/measurement/src/lib.rs
ginglis13: Good point, previous approach was to just log a warning here. New revision will fail on bad values.

ginglis13:
^ build failing due to unrelated lint - https://github.com/bottlerocket-os/bottlerocket-settings-sdk/pull/141 includes the fix, will re-run CI after that merges

piyush-jena (approved):

ginglis13:
^ force push to rebase for a clippy fix from #141

bottlerocket-settings-models/settings-extensions/measurement/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+/// Check whether a normalized exclusion is a self-reference to the measurement namespace.
+fn is_self_reference(normalized: &str) -> bool {
+    normalized == "measurement" || normalized.starts_with("measurement.")
+}
+
+/// Validate that a single exclusion pattern is allowed.
+fn validate_exclusion(value: &str) -> std::result::Result<(), MeasurementValidationError> {
+    let trimmed = value.trim();
+    let normalized = normalize_exclusion(trimmed);
+    if DISALLOWED_PATTERNS.contains(&normalized) {
```
piyush-jena: more rusty way to do it using ensure!()

piyush-jena: this applies to all 3 if conditions below.

piyush-jena: its used everywhere in our code base so you can use that ref

ginglis13:
^ force push uses more idiomatic `ensure!` macro for error conditionals

piyush-jena (approved):

bottlerocket-settings-models/settings-extensions/measurement/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+//!
+//! The `settings.measurement` namespace lets customers configure which settings
+//! are excluded from PCR 8 measurement by rottweiler.
+use bottlerocket_model_derive::model;
+use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
+use snafu::{ensure, Snafu};
+
+/// Patterns that must never be allowed as exclusions because they would
+/// undermine the integrity of the measurement system.
+const DISALLOWED_PATTERNS: &[&str] = &["settings", "", "*"];
```
Sparksssj: The PR description says validation rejects "wildcards," but this only matches a bare "*". Partial globs like "host-containers.*" or "*.foo" won't get rejected right? Is this expected or not?

ginglis13: `apiclient --exclude` is already a recursive prefix match. So to exclude everything under host-containers you write `host-containers`, and the subtree goes with it. host-containers.* / *.foo aren't rejected because they match a literal * segment that no key has, so they exclude nothing. 

You bring up a good point tho; allowing e.g. `host-containers.*` but silently having this no-op is not a great experience. I'll update to reject any `*` value.

ginglis13:
^ force push rejects any `*` appearing in any value of the excluded-settings list. Add an error message to make this clear, also update lib.rs docstring to clarify exclusion of top-level settings.

Sparksssj (approved):

piyush-jena (approved):
good call outs ^. LGTM

